PR #10, opened 2022-07-26
Update stellar-token-contract and stellar-contract-sdk

leighmcculloch:


leighmcculloch:
The number of changes I had to make to do the upgrade was much smaller than I anticipated... I guess these contracts don't use many of the things that got changed, which is super interesting.

PR #11, opened 2022-07-27
Update stellar-token-contract and stellar-contract-sdk

leighmcculloch:


liquidity_pool/src/lib.rs
```diff
@@ -166,8 +166,11 @@ impl LiquidityPoolTrait for LiquidityPool {
         let share_contract_id = create_contract(&e, &token_a, &token_b);
         let mut args = Vec::new(&e);
         args.push(get_contract_id(&e).into_env_val(&e));
-        e.call::<()>(
-            share_contract_id.clone().into(),
+        args.push(7u32.into_env_val(&e)); // TODO: Set decimals.
+        args.push(Binary::from_slice(&e, b"name").into()); // TODO: Set name.
+        args.push(Binary::from_slice(&e, b"symbol").into()); // TODO: Set symbol.
```
leighmcculloch, line 171: @jonjove @sisuresh In these two examples there were some new parameters to pass for the token contract, but I wasn't sure how they should be integrated, so I hardcoded some values that work with some TODOs for the moment.

PR #19, opened 2022-07-27
Add hello world example

leighmcculloch:


leighmcculloch:
@jonjove I added a simple example, a hello world, that uses types, and contract data. Could you take a look at make recommendations on if any of this can be simplified?

PR #46, opened 2022-08-01
Make it possible to test using the token contract WASM

### LINKED ISSUE #29: Add tests to examples that run the token contract with its WASM file
### What problem does your feature solve?

Right now, tests never run the WASM. That makes it hard to know if the WASM is up to date and actually works.

### What would you like to see?

Add a feature to run with the actual WASM.

### What alternatives are there?

TODO

jonjove:
### What

Adds new feature "token-wasm" that enables using the WASM. Adds a new line to Makefile to actually run this test.

### Why

Resolves #29.

### Known limitations

Doesn't pass tests right now because the WASM itself doesn't work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,6 +170,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "downcast-rs"
+version = "1.2.0"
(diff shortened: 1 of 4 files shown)


jonjove:
Closing this to change the target branch.

PR #59, opened 2022-08-11
Remove self-circular testutils deps

leighmcculloch:
### What
Remove self-circular testutils dev dependencies.

### Why

It causes errors in rust-analyzer which doesn't expect the circular reference. It is also reasonably confusing to others because (e.g. @graydon had questions about it) and requires a decent amount of discussion to knowledge share why it exists.

The circular reference was added to activate things that are gated behind `testutils` feature, so that those things are always activated during tests. However, Rust provides a built-in way for us to do this today, without needing the circular reference. All we need to do is use the `#[cfg(test)]` attribute. If we have items that need to be testutils for externals and we want them to always be activated within the crates own tests, then we can use `#[cfg(any(test, feature = "testutils"))]`. We already use this pattern in the SDK macros for any `testutils`.

See https://github.com/stellar/rs-soroban-sdk/pull/401

### DIFF
--- a/liquidity_pool/src/token_contract.rs
+++ b/liquidity_pool/src/token_contract.rs
@@ -1,10 +1,10 @@
 use soroban_sdk::{Binary, Env, FixedBinary};
 use soroban_token_contract::public_types::U256;
 
-#[cfg(any(not(feature = "testutils"), feature = "token-wasm"))]
+#[cfg(any(not(any(test, feature = "testutils")), feature = "token-wasm"))]
 pub const TOKEN_CONTRACT: &[u8] = include_bytes!("../../soroban_token_contract.wasm");
 
-#[cfg(any(not(feature = "testutils"), feature = "token-wasm"))]
+#[cfg(any(not(any(test, feature = "testutils")), feature = "token-wasm"))]
 pub fn create_contract(e: &Env, token_a: &U256, token_b: &U256) -> FixedBinary<32> {
     let bin = Binary::from_slice(e, TOKEN_CONTRACT);
     let mut salt = Binary::new(&e);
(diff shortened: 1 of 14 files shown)


liquidity_pool/src/token_contract.rs
```diff
@@ -1,10 +1,10 @@
 use soroban_sdk::{Binary, Env, FixedBinary};
 use soroban_token_contract::public_types::U256;
 
-#[cfg(any(not(feature = "testutils"), feature = "token-wasm"))]
+#[cfg(any(not(any(test, feature = "testutils")), feature = "token-wasm"))]
 pub const TOKEN_CONTRACT: &[u8] = include_bytes!("../../soroban_token_contract.wasm");
 
-#[cfg(any(not(feature = "testutils"), feature = "token-wasm"))]
+#[cfg(any(not(any(test, feature = "testutils")), feature = "token-wasm"))]
```
leighmcculloch, line 7: @jonjove This was pretty confusing to understand initially, so please let me know if I'm misunderstand how this is supposed to work.

I think the `testutils` cyclic dev-dependency hack was preventing any of the `token-wasm` feature usage to actually get used in tests, because tests always required testutils, and so there's no way `TOKEN_CONTRACT` could have been used in a test.

I'm not fixing that in this PR, I'm just filling in test gates for anything that was using testutils as a gate for whether tests were running or not. So I think a fix is needed if the hope is that `TOKEN_CONTRACT` will be used in tests.

leighmcculloch: Ah, I think I see how this works now, and I'm updating this incorrectly, will update it.

leighmcculloch: Ok, so no bug existed and I pushed a fix to this PR that I think updates the feature selection correctly. Still highly confusing.

jonjove: I think what you did is right. But I 100% agree that this is highly confusing. It was confusing when I wrote this, and it was confusing when I just reviewed these changes. Maybe there's a way to make this simpler. Maybe even just a comment would do the job.

leighmcculloch: I'll add a comment as part of this PR.

leighmcculloch: See b24f930.

jonjove (approved):

PR #62, opened 2022-08-15
Use native token contract in liquidity_pool

jonjove:
### What

Bump dependencies to pick up native contract support, use native token in liquidity_pool.

### Why

Native contracts are one of our goals.

### Known limitations

None

### DIFF
--- a/liquidity_pool/src/token_contract.rs
+++ b/liquidity_pool/src/token_contract.rs
@@ -2,58 +2,10 @@
 use soroban_sdk::{Bytes, BytesN, Env};
 use soroban_token_contract::public_types::U256;
 
-// Creating the token contract happens a couple different ways depending on the
-// situation:
-//
-// In tests, or when imported with testutils, without the token-wasm
-// feature, we use the imported token contract library and register it manually
-// as a test contract.
-//
-// In tests, when token-wasm feature is enabled, we use the embedded token wasm
-// file.
-//
-// Outside of tests and testutils, we use the embedded token wasm file.
-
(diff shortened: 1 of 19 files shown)


liquidity_pool/src/token_contract.rs
```diff
@@ -1,59 +1,11 @@
 #![allow(unused)]
-use soroban_sdk::{Binary, Env, FixedBinary};
+use soroban_sdk::{Bytes, BytesN, Env};
 use soroban_token_contract::public_types::U256;
```
leighmcculloch, line 3: The liquidity pool is still using the `soroban_token_contract`'s public types, rather than the native contracts types. I assume this is fine because they are identical.

What's the story do you think for how a contract gets these types? Will it be a separate crate in the `rs-stellar-env` repo that provides the types, and that we reexport in the SDK?

Note that the general way contract types will be imported into contracts will be via contract specs and codegen.

jonjove: Yes, this is correct. I've been thinking about how a contract should get these types. While providing those types in a crate is possible, I didn't go that route because I think it would be better to use the contract specs mechanism (this would make it possible to use all of the same machinery for native contracts). I intend to revisit this problem once the contract specs are more developed. I'll open an issue to track this.

leighmcculloch: > this would make it possible to use all of the same machinery for native contracts

I'm a few days away from this machinery being usable I think, so lets sync up then.

PR #79, opened 2022-08-25
Check if rust-analyzer can parse project

leighmcculloch:
### What
Add ci job that checks if rust-analyzer can parse project.

### Why
A couple times now we have done things that are technically allowed in Rust but are not supported by rust-analyzer. Everytime we've walked the change back because rust-analyzer compatibility is important for developer productivity. Let's prevent these things from happening by checking if rust-analyzer can parse the project.

Same as https://github.com/stellar/rs-soroban-env/pull/371.
### Known Limitations

This relies on parsing the output of rust-analyzer analysis-status which is not stable, so it may change or break over time (https://github.com/rust-lang/rust-analyzer/issues/13115). We can remove this check if it doesn't live up to its hoped impact.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,7 @@ jobs:
 
   complete:
     if: always()
-    needs: [fmt, build-and-test]
+    needs: [fmt, rust-analyzer-compat, build-and-test]
     runs-on: ubuntu-latest
     steps:
     - if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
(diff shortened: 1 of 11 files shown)


leighmcculloch:
I just went through all the examples and remove the cyclic self-references. The changes are in this PR.

We had removed these previously in #59, but looks like they crept back into some new examples we wrote.

For anyone adding examples here (@sisuresh @jonjove?), can you take a look at the diff to see what we're doing now? Instead of adding a self-referencing dev dependency with `testutils` feature enabled, we add a `cfg(test)` everywhere as an alternative way to trigger any `testutils` feature gated code, that way testutils functionality is just always available in tests.

PR #89, opened 2022-08-30
Use contractimport! in liquidity_pool examples

leighmcculloch:
### What
Use `contractimport!` in `liquidity_pool` example to import WASM file, and to import client.

### Why
Start updating contracts to use the contractimport mechanics, and to move away from importing each other's Rust crates. Lean in on using the WASM for testing.

This doesn't go all the way to using all of the contractimport machinery, because it still relies on the token contract's utilities for native testing, and the utilities for signing and nonce management in tests as well. These things will be follow up looked at in a future iteration in:
- https://github.com/stellar/rs-soroban-sdk/issues/528
- https://github.com/stellar/rs-soroban-sdk/issues/529


liquidity_pool/src/test.rs
```diff
@@ -32,7 +33,8 @@ fn generate_keypair() -> Keypair {
 }
 
 fn create_token_contract(e: &Env, id: &[u8; 32], admin: &Keypair) -> Token {
-    register_token(&e, id);
+    let contract_id = BytesN::from_array(e, &id);
+    e.register_contract_wasm(&contract_id, token_contract::WASM);
     let token = Token::new(e, id);
     // decimals, name, symbol don't matter in tests
     token.initialize(&to_ed25519(&e, admin), 7, "name", "symbol");
```
leighmcculloch, line 40: I'd like to replace `token`, i.e. `soroban_token_contract::testutils::Token`, with the generated client that is in `token_contract::ContractClient`. Right now the former has a combination of helpers for calling a contract, as well as helpers for producing the right types of messages for those calls. I think we should split that in half, so that the only utilities a consuming contract needs to import as code is those utilities, not the calling functions.

We can then assess whether those utilities can be generalized into the SDKs at all.

leighmcculloch (commented):
@sisuresh @jonjove I started work on updating the examples to use `contractimport!` and the generated contract client on the calling side, but I only got part way. I did the non-test side, and I did the deployment of the contract on the test side, but I didn't do the helper Token class. I think we probably need to collaborate on that. See below 👇🏻

liquidity_pool/src/test.rs
jonjove: I'm pretty unclear on what you actually want to change here. Can you pick a specific function (eg `soroban_token_contract::testutils::Token::approve`) and explain what you want to be different in terms of implementation/usage/API?

leighmcculloch: We discussed and going to revisit this. We can't do this now because there's too much convenience logic in the Token type that isn't included in the contract client, but @jonjove has some ideas about how we could generate additional clients with that functionality.

Issue #73, opened 2022-08-22
Add smart wallet example


jonjove:
### What problem does your feature solve?

Although smart wallets aren't widely used, they are a possible construction on Soroban and it would be good to know how they fit into the system.

### What would you like to see?

A simple smart wallet example. The most basic would be just an on-chain multi-signature configuration analogous to classic Stellar accounts.

### What alternatives are there?

N/A

jonjove:
Another option would be to make a simple example similar to gnosis-safe that can take signatures asynchronously (eg. someone proposes a transaction and other signers confirm it). See https://docs.gnosis-safe.io/introduction/the-programmable-account/eoas-vs.-contract-accounts#multi-signature-wallets for context.

This is more complex, but has the nice feature of being something that cannot be done using classic Stellar accounts.

tomerweller:
Another aspect of smart wallets that might be worth exploring is velocity controls. This also has the bonus of being not trivial in Stellar classic. 

A simple version is that there's a single admin signer with a fixed allowance per calendar day. More complex versions might include (1) combining with multisig for tiered limits based on number of signers. and/or (2) using sliding windows instead of calendar days

smac711:
@jonjove @dmkozh are we actively working on this issue. If not and it is not critical or completion I will be moving it out of the 22.09.01 sprint to backlog.

jonjove:
Yes #106 is open and will be merged in the next few days.

smac711:
Hey @dmkozh are we ready to merge the above pull request in order to close off this issue?

leighmcculloch:
I moved this into the Oct iteration and marked it as part of the Test & Polish steel thread. Anything in this repo is unlikely to classify as a code complete requirement.

PR #106, opened 2022-09-13
Example simple wallet with async multisig support.

dmkozh:
### What

Simple smart wallet example (https://github.com/stellar/soroban-examples/issues/73).

The wallet supports weighted multisig with arbitrarily separated signature batches (i.e. the payment may happen both immediately or after a sequence of confirmations from signers). 

### Why

With this example I've tried to build multisig contract using the existing SDK tools. This should be useful as we keep iterating on auth, signing utils etc.

### Known limitations

Some book-keeping (like payment id generation) is not implemented in the contract itself, though that doesn't prevent it from being correct/functional.


wallet/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+        read_nonce(&env, admin)
+    }
+
+    // Stores a provided payment or executes it when enough signer weight is
+    // accumulated.
+    // Returns `true` when the payment was executed and `false` otherwise.
+    //
+    // Every wallet admin signs `pay` as if it was called by them only, i.e.
+    // they should sign `pay` function call with argument tuple of
+    // `(id, nonce, payment_id, payment)`. Then the signatures and nonces of
```
leighmcculloch, line 113: We should encourage including a domain separator in anything that gets signed, so we should probably add it to this example, if we can find a concise way to do so.

wallet/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+crate-type = ["cdylib", "rlib"]
+doctest = false
+
+[features]
+testutils = ["soroban-sdk/testutils", "soroban-token-contract/testutils"]
+
+[dependencies]
+soroban-sdk = "0.0.4"
+soroban-auth = "0.0.4"
+soroban-token-contract = { git = "https://github.com/stellar/soroban-token-contract", rev = "46ca69b8", default-features = false }
```
leighmcculloch, line 19: Before we can merge this contract we need to update how it interacts with tokens by importing the `soroban-token-spec.wasm` for the client, and creating and using the built-in token contract.

leighmcculloch (commented):
This is very nice example. For something that has a certain depth of complexity, the example presents that complexity in a way that is rather brilliant in its simplicity.

There's one thing I mention inline I think we should tweak so as to encourage a good security practice. There's one thing I mention inline we need to update just because it is out-of-date.

I'm happy to help update the contract, but defer to you @dmkozh if you want to do it.

cc @tomerweller if you haven't seen this contract, it is worth a read.

wallet/Cargo.toml
dmkozh: Updated the contract to match the new SDK and token contract.

wallet/src/lib.rs
dmkozh: I'm not sure what do you mean by the domain separator here. Network id is a part of the signature payload expected by `verify`. Do you mean something else here?

leighmcculloch: Ok, I missed that this section was using soroban-auth. As long as it's using the signature payload from that crate then it's good.

wallet/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+            env.data().set(DataKey::Payment(payment_id), payment);
+        }
+
+        false
+    }
+}
+
+fn check_initialization_params(env: &Env, admins: &Vec<Admin>, threshold: u32) {
+    if threshold == 0 {
+        panic!("threshold has to be non-zero");
```
leighmcculloch, line 178: Given the amount of input validation this contract would benefit from using `panic_error!` I think, so that if there are panics in the final compiled, there are contract error codes rather than the trap error code. See https://soroban-dev.stellar.org/docs/examples/errors#panicking-with-an-error.

leighmcculloch (approved):

wallet/src/lib.rs
dmkozh: I agree that errors would be nice. Moreover, I would probably rewrite this to return `Result` instead of panicking and use `try_` functions in tests as `should_panic` is pretty annoying. The only thing I'm not sure about is granularity: replacing every panic with a unique error is a bit annoying (and also inconvenient for a potential consumer of these errors), while reducing granularity would lose some information/test coverage (now I can precisely set expectations on why should the contract fail). So ideally I'd want to both return an error and have a string explanation (also as a debug event?) for debugging/testing.

leighmcculloch: I'd lean into granularity.

You can still use the `try` functions if you use the `panic_error!` macro. This might be a good balance if the intent is mostly for testing and diagnostics, and less for developers to handle them explicitly. The only difference between returning Result or not is that the Error gets encoded into the contract spec. So if you want people to handle errors, getting it into the contract spec helps to document it.

> So ideally I'd want to both return an error and have a string explanation (also as a debug event?) for debugging/testing.

We might want to explore modifying `panic_error!` so that it can optionally take a string as well, and it can log a debug event using `log!` alongside storing the error.

dmkozh: > I'd lean into granularity.

True, I guess unlike with host more granular errors are more acceptable for contracts (given that they're immutable). 

> We might want to explore modifying panic_error! so that it can optionally take a string as well, and it can log a debug event using log! alongside storing the error.

That's what I've been thinking about, however currently the debug events from `log` don't seem to appear in the test logs, so we need to think how to make this useful. For now I'll just use granular errors.

leighmcculloch: You can make the logs appear in test logs by adding this line to the end of your test:
```rust
env.logger().print()
```

I'd love to find a way to hook this in automatically if you have any idea.

dmkozh: > The only difference between returning Result or not is that the Error gets encoded into the contract spec.

It seems like another different thing is that `panic_error` still panics, which makes tests harder (and IIRC we can't `catch_unwind` that). So for the sake of testing `Result` seems preferential, as `should_panic` is inconvenient when there are a lot of failure scenarios to cover.

leighmcculloch: > `panic_error` still panics, which makes tests harder

Agreed, and there's an issue open to address this: 
- https://github.com/stellar/rs-soroban-sdk/issues/642
- https://github.com/stellar/rs-soroban-env/issues/430

leighmcculloch: Fwiw I think returning Results from these functions would also be very reasonable. I'm really curious if you add that how much the contract size changes. I think we've been worried return paths could be expensive, but it's all assumptions at this point, and you might prove that wrong, which imo would be awesome. So awesome we could maybe even axe panic_error!.

dmkozh: In https://github.com/stellar/soroban-examples/pull/150 I'm using results everywhere besides `ConversionError` unwraps. I think using result doesn't make things much more complex/verbose, It definitely is beneficial for testing and the code is IMHO more idiomatic Rust.

PR #120, opened 2022-09-21
Payload exploration

jonjove:
### What

This PR contains a still-in-progress exploration of having contracts expose a `payload` function to facilitate signing in wallets and other downstream systems. To make development as fast as possible, I've replicated/modified auth functionality in this crate rather than working in soroban-auth. The final implementation of course would be in soroban-auth.

Earlier discussions of this idea used names like `f_payload` but in general those names don't comply with the 10 character symbol requirement. The universal `payload` function is the reaction to that issue. The signature of `payload` is almost certain to change as this exploration develops.

This also incorporates an idea from @dmkozh about putting all signatures into a single argument at the end of the argument list. Exploration on this part is very preliminary, and is almost certain to change entirely.

cc @leighmcculloch who requested this PR

### Why

Making signing easier is a huge piece of our work in the last several weeks.

### Known limitations

Too many to enumerate

### DIFF
--- a/liquidity_pool_router_with_payload/src/lib.rs
+++ b/liquidity_pool_router_with_payload/src/lib.rs
@@ -160,7 +160,7 @@ pub trait PayloadTrait {
         function: Symbol,
         args: Vec<RawVal>,
         callstack: Vec<(BytesN<32>, Symbol)>,
-    ) -> Map<Symbol, SaltedSignaturePayload>;
+    ) -> Map<Symbol, (Identifier, SaltedSignaturePayload)>;
 }
 
 pub struct LiquidityPoolRouterPayload;
@@ -172,7 +172,7 @@ impl PayloadTrait for LiquidityPoolRouterPayload {
         function: Symbol,
         args: Vec<RawVal>,
         callstack: Vec<(BytesN<32>, Symbol)>,
-    ) -> Map<Symbol, SaltedSignaturePayload> {
+    ) -> Map<Symbol, (Identifier, SaltedSignaturePayload)> {
         const DEPOSIT_RAW: u64 = symbol!("deposit").to_raw().get_payload();
         match function.to_raw().get_payload() {
             DEPOSIT_RAW => {
@@ -223,7 +223,7 @@ impl PayloadTrait for LiquidityPoolRouterPayload {
                 let mut res = Map::new(&e);
                 res.set(symbol!("xfer_a"), to_forward_a);
                 res.set(symbol!("xfer_b"), to_forward_b);
-                res.set(symbol!("deposit"), to_verify);
+                res.set(symbol!("deposit"), (to, to_verify));
             }
             _ => panic!(),
         }
--- a/liquidity_pool_router_with_payload/src/lib.rs
+++ b/liquidity_pool_router_with_payload/src/lib.rs
@@ -1,0 +1,72 @@
+#![no_std]
+
+mod auth;
+
+use crate::auth::{PayloadTrait, SignaturePayload, SignaturePayloadV0};
+use soroban_auth::{Identifier, Signature};
+use soroban_sdk::{contractimpl, contracttype, BigInt, BytesN, Env, Map, RawVal, Symbol, Vec};
+
+fn get_deposit_amounts(
+    desired_a: BigInt,
+    min_a: BigInt,
+    desired_b: BigInt,
+    min_b: BigInt,
+    reserves: (BigInt, BigInt),
+) -> (BigInt, BigInt) {
+    if reserves.0 == 0 && reserves.1 == 0 {
+        return (desired_a, desired_b);
+    }
+
+    let amount_b = desired_a.clone() * reserves.1.clone() / reserves.0.clone();
+    if amount_b <= desired_b {
+        if amount_b < min_b {
+            panic!("amount_b less than min")
+        }
+        return (desired_a, amount_b);
+    } else {
+        let amount_a = desired_b.clone() * reserves.0 / reserves.1;
+        if amount_a > desired_a || desired_a < min_a {
+            panic!("amount_a invalid")
+        }
+        return (amount_a, desired_b);
+    }
+}
+
+#[contracttype]
+pub struct Deposit {
+    pub token: BytesN<32>,
+    pub nonce: BigInt,
+    pub desired: BigInt,
+    pub min: BigInt,
+}
+
+pub trait LiquidityPoolRouterTrait {
+    fn deposit(
+        e: Env,
+        to: Identifier,
+        deposit_a: Deposit,
+        deposit_b: Deposit,
+        sigs: Map<Identifier, Signature>,
+    );
+}
+
+pub struct LiquidityPoolRouterPayload;
+
+#[cfg_attr(feature = "export", contractimpl)]
+#[cfg_attr(not(feature = "export"), contractimpl(export = false))]
+impl PayloadTrait for LiquidityPoolRouterPayload {
+    fn payload(
+        e: Env,
+        function: Symbol,
+        args: Vec<RawVal>,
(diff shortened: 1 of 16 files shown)


token_with_payload/src/contract.rs
```diff
@@ -0,0 +1,290 @@
...
+        nonce: BigInt,
+        spender: Identifier,
+        amount: BigInt,
+        sigs: Map<Identifier, Signature>,
+    ) {
+        let from = sigs.get(from_id.clone()).unwrap().unwrap();
+
+        verify_and_consume_nonce(&e, &from_id, &nonce);
+
+        crate::auth::check_auth::<TokenPayload>(
```
dmkozh, line 128: Is the idea to use/reuse different payloads for different functions that need auth? I suppose you omitted that from the PR, as only this fn uses the new auth.

token_with_payload/src/contract.rs
```diff
@@ -0,0 +1,290 @@
...
+
+pub struct TokenPayload;
+
+#[cfg_attr(feature = "export", contractimpl)]
+#[cfg_attr(not(feature = "export"), contractimpl(export = false))]
+impl PayloadTrait for TokenPayload {
+    fn payload(
+        e: Env,
+        function: Symbol,
+        args: Vec<RawVal>,
```
dmkozh, line 279: For me the main value of the initial proposal for having `fn_payload` was that we have a typed interface to contract fn we need to sign. Here it's still not clear what the args should be, which I think is on of the main issues with signing fns. Or is this complimentary to the initial proposal and contracts still would introduce fn_payload for fns with non-trivial args mapping?

token_with_payload/src/contract.rs
jonjove: It's exploratory work. I learned what I needed from this part so I didn't do the rest of the work (as noted, I expect everything to change). But in practice every function will use the same new auth.

token_with_payload/src/contract.rs
jonjove: There's not really any easy way to do the `f_payload` thing as noted in the PR description. I agree typed interfaces are nice, and I would prefer that. For the sake of this exploration, I went this way because it meant I could move fast without solving other problems (like length limits). The real implementation might look different.

token_with_payload/src/contract.rs
```diff
@@ -0,0 +1,290 @@
...
+    if nonce != expected_nonce {
+        panic!("incorrect nonce")
+    }
+    e.contract_data().set(key, &nonce + 1);
+}
+
+pub struct Token;
+
+#[cfg_attr(feature = "export", contractimpl)]
+#[cfg_attr(not(feature = "export"), contractimpl(export = false))]
```
leighmcculloch, line 95: Just an fyi, this doesn't do anything anymore. `contractimpl` always exports.

jonjove: Good to know. I copied and pasted this from the token example, so it should be removed there too.

jonjove:
Added the missing pieces described above and fixed a couple of bugs. A wallet or other downstream system would use this as follows:

1. Wallet receives request to call `Contract::f(x, y, z...)`
2. Wallet calls `Contract::has_sig(f)`. If false, stop here and just submit the request as received. If true, continue to step 3.
3. Wallet checks that the last parameter to `f` is an empty map (there might be other things that could be done at this step but this is reasonable for starters)
4. Wallet calls `Contract::payload(f, [args excluding the last parameter which is an empty map], [])`
5. Wallet takes the result of `payload` and requests the appropriate signatures using the `Identifier` and `SaltedSignaturePayload` in each value, then packages those signatures back into a map using the `Symbol` in each key
6. Wallet submits `Contract::f(x, y, z...)` with the last parameter (which was an empty map) replaced by the map of signatures produced in step 5

I know that this flow is pretty complex, but I think it covers an enormous number of cases including interesting cases that involve functions that forward some-but-not-all signatures to other functions. See liquidity_pool_router_with_payload `deposit` for context on how that works.

This is basically concept-complete at this point so feedback on the idea would be great. Sorry that the implementation is largely hideous, but I'm confident it could be cleaned up.

jonjove:
I forgot to mention that the wallet needs to do some verification on the payloads it receives before requesting signatures. Let me think about exactly what needs to be checked.

liquidity_pool_router_with_payload/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+                let amounts = get_deposit_amounts(
+                    deposit_a.desired,
+                    deposit_a.min,
+                    deposit_b.desired,
+                    deposit_b.min,
+                    reserves,
+                );
+
+                let mut callstack_a = callstack.clone();
+                callstack_a.push_back((deposit_a.token.clone(), symbol!("xfer")));
```
dmkozh, line 204: Should call stack a contain this contract (i.e. liquidity pool)?

liquidity_pool_router_with_payload/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+            &Identifier::Contract(e.get_current_contract()),
+            &amounts.1,
+            &sigs_b,
+        );
+
+        pool_client.deposit(&to);
+    }
+}
+
+pub trait PayloadTrait {
```
dmkozh, line 157: Would this trait be a part of a library? I don't see it used in auth now.

liquidity_pool_router_with_payload/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+            _ => false,
+        }
+    }
+
+    fn payload(
+        e: Env,
+        function: Symbol,
+        args: Vec<RawVal>,
+        callstack: Vec<(BytesN<32>, Symbol)>,
+    ) -> Map<Symbol, (Identifier, SaltedSignaturePayload)> {
```
dmkozh, line 185: This looks nice, pretty close to what I've imagined, albeit a bit verbose. Maybe we could provide a simple implementation for single-signature contracts, while allowing more complex contracts to be flexible.

jonjove: I'm very open to having multiple versions of this, and I actually think that's one of the major strengths of this approach. For example, we could add a function `fn payload_v(e: Env) -> Symbol` (`payload_v` is short for "payload version") that returns an identifier for what payload function to call. We could even make that `fn payload_v(e: Env, function: Symbol) -> Symbol` if we wanted to allow it to vary by function. Then we could have a couple standards at different complexity levels, and the wallet would just invoke the right one based on the return value of `payload_v`.

liquidity_pool_router_with_payload/src/lib.rs
jonjove: Great question. I think it could go either way. Contracts would probably be mildly cheaper if we defined "call stack" to mean everything before this function--this means the simple case of passing a signature to the first contract incurs no call stack cost since it would be empty. But the whole system would probably be slightly less surprising if we defined "call stack" to mean everything including this function. What do you think?

liquidity_pool_router_with_payload/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+                let to_forward_b = token_contract::Client::new(&e, &deposit_b.token)
+                    .payload(
+                        &symbol!("xfer"),
+                        &(&to, deposit_b.nonce, e.get_current_contract(), amounts.1).into_val(&e),
+                        &callstack_b,
+                    )
+                    .get_unchecked(symbol!("sig"))
+                    .unwrap();
+
+                let to_verify = SaltedSignaturePayload::V0(SaltedSignaturePayloadV0 {
```
jonjove, line 225: I think this isn't quite right. I'm pretty sure I would need to add `to_forward_a` and `to_forward_b` to this signature payload.

bp2m6v4sp5-dev (approved):

Issue #146, opened 2022-10-09
Add simple token usage example


tomerweller:
Our current examples that make use of tokens cover a lot of ground. Single offer, Liquidity pool and Timelock are all great but are not trivial and also implement advanced auth.   

It would be beneficial to have an almost trivial contract example accompanied by a tutorial that focuses on basic token usage. 

One option is to make a simplified version of the timelock contract: single claimant, only "after" timebound, no advanced auth.

tomerweller:
@dmkozh, as the authoer of timelock, is simplifying it a low hanging fruit?

leighmcculloch:
I think we could remove the use of soroban-auth from `timelock`, because it takes on a separate call to token approve anyway.

dmkozh:
FWIW for `token_transfer` we could get rid of the time lock too, just have deposit to contract and define a single claimant, then in claim just check for the claimant only. This way the example would be focused on transfers too/from contract only.

leighmcculloch:
> so that we could just use the invoker and not require an advance approve call, right?

You'd still need the advanced approve call I think.

The rest of the example, unrelated to auth, seems fine to me.

leighmcculloch:
> we should probably consider passing the source account recursively to all the sub-contract calls (unless there is some security risk, but I don't see it)

This would open up a significant security footgun opportunity: relay attacks. This is where a message sent to contract A can be relayed to contract B, even if the invoker didn't intend.

Ethereum provides a way to find out the original invoker via the `tx.origin`, but it should really never be used in authorization. E.g. https://hackernoon.com/hacking-solidity-contracts-using-txorigin-for-authorization-are-vulnerable-to-phishing

PR #148, opened 2022-10-10
Simplified auth in timelock example.

dmkozh:
### What

Now the example only uses invoker auth.

### Why

As discussed in https://github.com/stellar/soroban-examples/issues/146, the auth is not that relevant for this example, 

### Known limitations

N/A


leighmcculloch (approved):
Nice, thanks for updating.

PR #149, opened 2022-10-11
Use the SDK conversion from address to identifier.

dmkozh:
### What

Use the SDK conversion from address to identifier.

I'm ok with postponing this if we don't want to bump the SDK revision, this is mostly to not forget about the change.

### Why

Simplifying example code

### Known limitations

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,7 +646,7 @@ dependencies = [
 [[package]]
 name = "soroban-auth"
 version = "0.1.0"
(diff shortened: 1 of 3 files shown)


leighmcculloch:
> I'm ok with postponing this if we don't want to bump the SDK revision, this is mostly to not forget about the change.

Let's merge it. If we need to tag a new release, we can figure it out. The examples referenced by the docs reference a tag on this repo too fyi.

PR #177, opened 2022-11-29
Update the examples to reflect the contract deployment changes

dmkozh:
### What

Update the SDK revision and change the examples to deploy the contracts using the WASM hash, as well as tests to not use contract ids.

### Why

Support the SDK changes.

### Known limitations

N/A

### DIFF
--- a/deployer/deployer/src/test.rs
+++ b/deployer/deployer/src/test.rs
@@ -15,9 +15,11 @@ fn test() {
     let env = Env::default();
     let client = DeployerClient::new(&env, &env.register_contract(None, Deployer));
 
-    // Deploy contract using deployer, and include an init function to call.
-    let salt = Bytes::from_array(&env, &[0; 32]);
+    // Install the WASM code to be deployed from the deployer contract.
     let wasm_hash = env.install_contract_wasm(contract::WASM);
+    
+    // Deploy contract using deployer, and include an init function to call.
+    let salt = Bytes::from_array(&env, &[0; 32]);    
     let init_fn = symbol!("init");
     let init_fn_args = (5u32,).into_val(&env);
     let (contract_id, init_result) = client.deploy(&salt, &wasm_hash, &init_fn, &init_fn_args);
--- a/deployer/deployer/src/test.rs
+++ b/deployer/deployer/src/test.rs
@@ -1,7 +1,7 @@
 #![cfg(test)]
 
 use crate::{Deployer, DeployerClient};
-use soroban_sdk::{bytesn, symbol, Bytes, BytesN, Env, IntoVal};
+use soroban_sdk::{symbol, Bytes, Env, IntoVal};
 
 // The contract that will be deployed by the deployer contract.
 mod contract {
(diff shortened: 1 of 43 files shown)


deployer/deployer/src/test.rs
```diff
@@ -13,20 +13,14 @@ mod contract {
...
     let env = Env::default();
-    let deployer_id = BytesN::from_array(&env, &[0; 32]);
-    env.register_contract(&deployer_id, Deployer);
-    let client = DeployerClient::new(&env, &deployer_id);
+    let client = DeployerClient::new(&env, &env.register_contract(None, Deployer));
 
     // Deploy contract using deployer, and include an init function to call.
     let salt = Bytes::from_array(&env, &[0; 32]);
-    let wasm: Bytes = contract::WASM.into_val(&env);
+    let wasm_hash = env.install_contract_wasm(contract::WASM);
```
leighmcculloch, line 20: Could you move this line above the comment, and add its own comment signaling that it is being installed prior to deployment? I think pulling it out will help people understand that step clearer.

deployer/deployer/src/test.rs
```diff
@@ -13,20 +13,14 @@ mod contract {
...
     let init_fn = symbol!("init");
     let init_fn_args = (5u32,).into_val(&env);
-    let (contract_id, init_result) = client.deploy(&salt, &wasm, &init_fn, &init_fn_args);
-    assert_eq!(
-        contract_id,
-        bytesn!(&env, 0xead19f55aec09bfcb555e09f230149ba7f72744a5fd639804ce1e934e8fe9c5d)
```
leighmcculloch, line 28: It's a shame we lose this fixed contract ID, because we had it fixed so that we can discuss the ID in the docs and people cna copy paste a command to invoke that contract. Is there anyway to keep it fixed?

leighmcculloch (approved):
One ask, but if it isn't possible lgtm as it is.

deployer/deployer/src/test.rs
leighmcculloch: Don't worry about this, I don't think it will compromise the docs that much. And if it does we can circle back on the example.

leighmcculloch: For reference this is the example: https://soroban.stellar.org/docs/examples/deployer

deployer/deployer/src/test.rs
dmkozh: Updated.

deployer/deployer/src/test.rs
dmkozh: I've actually been thinking about this one. There are actually a couple of issues that this highlights:

- The fixed contract ids - I've had the respective comment in the SDK PR. We may need to think about providing some functionality that wouldn't be too far from reality, while also convenient to use. That said, the deployer still can have a fixed contract id, I just removed it for consistency with other examples (and we also were going to remove it for the compiled contracts as well)
- The particular hardcoded ID here - I don't think it's useful enough here: even if I kept a stable deployer id, I still don't know how to update this value (we've changed the ID preimage hence it would be a constant, but different value). I also wouldn't have a way to verify the value correctness. What I think would be useful is to provide utilities to generate contract ids. E.g. here a more useful assertion would be `assert_eq!(SomeUtil::get_contract_id_from_contract(env.get_ledger_network_id(), parent_contract_id, salt), contract_id)`. This doesn't even require fixed/stable contract ids, while being easier to maintain and more expressive.

PR #200, opened 2023-02-03
Update to use rs-soroban-sdk 0.4.3

silence48:
### What

Use rs-soroban-sdk 0.4.3
update the gitpod dockerfile to have a futurenet rpc terminal and a sandbox.

### Why
compatability

### Known limitations

NA

<a href="https://gitpod.io/#https://github.com/stellar/soroban-examples/pull/200"></a>


.gitpod.yml
```diff
@@ -2,7 +2,32 @@ image:
...
+      stellar/quickstart:soroban-dev \
+      --futurenet \
+      --enable-soroban-rpc
+  # This "CLI - Futurenet" task opens a terminal for you to interact with the
+  # futurenet network.
+  - name: CLI - Futurenet
+    # We specify some required environment variables for use on the futurenet.
+    env:
+      # This might should be set by the user.
+      # This account must actually exist as a funded account on the futurenet.
```
dmkozh, line 22: Seems like this comment refers to something other than RPC URL

.gitpod.yml
```diff
@@ -2,7 +2,32 @@ image:
...
+      -p 8000:8000 \
+      stellar/quickstart:soroban-dev \
+      --futurenet \
+      --enable-soroban-rpc
+  # This "CLI - Futurenet" task opens a terminal for you to interact with the
+  # futurenet network.
+  - name: CLI - Futurenet
+    # We specify some required environment variables for use on the futurenet.
+    env:
+      # This might should be set by the user.
```
dmkozh, line 21: nit: might or should?

dmkozh (commented):
The dockerfile/gitpod changes look good, but the SDK version bump requires a pass through the examples to account for the interface changes (storage now uses references). I'm working right now on updating the examples to even more recent SDK version, so not sure if you'd want to do that for your PR.

.gitpod.yml
silence48: oops this was a copy and paste from another file I will fix it.

.gitpod.yml
silence48: That irrelevant comment has been removed thanks.

.gitpod.yml
silence48: I reworded line 21

silence48:
I solved the error in the tests by changing the sdk commit to be e1c3de3. they should be passing now.

silence48:
> The dockerfile/gitpod changes look good, but the SDK version bump requires a pass through the examples to account for the interface changes (storage now uses references). I'm working right now on updating the examples to even more recent SDK version, so not sure if you'd want to do that for your PR.

I changed it to use commit e1c3de3 of the sdk and all the tests pass now.

.gitpod.yml
```diff
@@ -2,7 +2,31 @@ image:
...
 
 tasks:
-  - init: |
+  - name: Futurenet
+    init: |
+      docker pull stellar/quickstart:soroban-dev
+    command: |
+      docker run --rm -it \
+      --name stellar \
+      --platform linux/amd64 \
```
sreuland, line 11: `stellar/quickstart:soroban-dev` is now published with multi-arch support, so, I don't think you need to force the amd64 platform here, should be able to remove it.

silence48: I removed the platform designation; Thank you.

dmkozh (approved):

PR #216, opened 2023-02-17
feat: add CLI tests

willemneal:
Currently the docs have example cli commands for the examples and this provides tests to ensure that the docs are accurate.

<a href="https://gitpod.io/#https://github.com/stellar/soroban-examples/pull/216"></a>


test/cli-tests/tests/it/auth.rs
```diff
@@ -0,0 +1,50 @@
+use soroban_test::{TestEnv, Wasm};
+
+const WASM: &Wasm = &Wasm::Release("soroban_auth_contract");
+
+const ACC1: &str = "GA6S566FD3EQDUNQ4IGSLXKW3TGVSTQW3TPHPGS7NWMCEIPBOKTNCSRU";
```
willemneal, line 5: Please write out.  No need to be so terse in an example test.

chadoh: done in d6bf4f8

paulbellamy (approved):

test/cli-tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+publish = false
+
+
+[lib]
+crate-type = ["cdylib"]
+doctest = false
+
+[dev-dependencies]
+predicates = "2.1.5"
+soroban-test = {  git = "https://github.com/ahalabs/soroban-tools", rev = "bc1d008" }
```
paulbellamy, line 16: This should probably be the stellar repo, right?

paulbellamy:
The tests themselves look fine. But they feel like cli tests, so it's odd they're in the examples repo. Feels like they should be in the `soroban-tools` repo.

tomerweller:
I agree with @paulbellamy that these tests don't naturally fit in the examples repo. I think that in a perfect world they would be in the docs repo, as the docs are what is actually being tested. However, the docs don't currently have machinery to facilitate that. So for now having them here sounds like a reasonable compromise to me.  

I suggest the following: 
1. rename `cli-tests` to `docs-tests` or `docs-cli-tests` to reflect the fact that we're testing docs, not the cli 
2. add a short README to this new tests package to explain what it does
3. add a rustdoc at the top of every test file that links to the relevant doc it's testing

Thoughts @willemneal @chadoh?

willemneal:
I think doc-tests is a good name for now.  Ideally we would make the examples a git submodule and then the docs repo could host the tests there.

tomerweller:
I do wonder if there's a more ergonomic way to write these tests. Constructing these commands in code is somewhat painful and will be painful to update. 

Is there a way to use the cli's parser to parse a single command string instead of constructing it by code? I think this will make writing and updating these tests easier - just copy&paste the cli command.

chadoh:
@tomerweller Willem and I would rather have the ability to move somewhat in the other direction. 

- https://github.com/stellar/soroban-tools/pull/443 

^ this would allow us to:

- depend on a specific version of the CLI
- write Rust tests that execute Rust code, rather than calling out to an external CLI process
- speed up the tests

What do you think?

test/cli-tests/Cargo.toml
chadoh: Yes, relevant PR is here:

- https://github.com/stellar/soroban-tools/pull/437

We can wait to merge these examples tests until the soroban-tools PR is merged.

chadoh:
> I suggest the following:
>
> - rename cli-tests to docs-tests or docs-cli-tests to reflect the fact that we're testing docs, not the cli
> - add a short README to this new tests package to explain what it does
> - add a rustdoc at the top of every test file that links to the relevant doc it's testing

All done.

Looks like someone needs to approve running the workflow so tests actually run.

tomerweller:
> write Rust tests that execute Rust code, rather than calling out to an external CLI process

I totally agree about writing tests in rust and not calling an external CLI process. 

My question/wonder was whether there's a way _in code_ to construct the cli command, using the library, instead of "manually" constructing it.

So instead of this

```rust 
TestEnv::with_default(|e| {
e.new_cmd("contract")
  .arg("invoke")
  .arg("--wasm")
  .arg(&WASM.path())
  .args(["--id", "1"])
  .args(["--fn", "increment"])
  .args(["--"])
  .args(["--incr", "5"])
  .assert()
  .stderr("")
  .stdout("5\n");
});
```

One could potentially write something more "similar" to this: 

```rust
TestEnv::with_default(|e| {
  e.new_cmd_pasre("contract invoke --wasm {} --id 1 --fn increment -- --incr 5", &WASM.path())
    .assert()
    .stderr("")
    .stdout("5\n");
});
```

This makes an assumption that there's an available parse command (or that one could be exposed) in the cli lib. Intuitively I think there should be one because the soroban-cli indeed parses commands from the cli that take that shape (but my intuition is often wrong). What's the feasibility of this?

willemneal:
The idea would be to allow constructing an invoke, or any command, and then call a run function that returns the output. 

We have tests in the cli to handle testing cli parsing but we can skip that part and have proper types.

test/doc-tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "cli-tests"
+version = "0.0.0"
+authors = ["Stellar Development Foundation <info@stellar.org>"]
+license = "Apache-2.0"
+edition = "2021"
```
tsachiherman, line 6: not an expert, but shouldn't this be 2023 ?

test/doc-tests/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+publish = false
+
+
+[lib]
+crate-type = ["cdylib"]
+doctest = false
+
+[dev-dependencies]
+predicates = "2.1.5"
+soroban-test = {  git = "https://github.com/ahalabs/soroban-tools", rev = "698479c" }
```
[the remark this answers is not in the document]
chadoh: Yes, see related discussion [from last week](https://github.com/stellar/soroban-examples/pull/216#pullrequestreview-1313969085)

paulbellamy (approved):
LGTM. It would be nice to have these run as part of CI

chadoh:
@paulbellamy can you approve the workflow to run?

paulbellamy:
Nope, I don't have write access to this repo. Maybe @leighmcculloch can?

tsachiherman:
> 

Just did. Could you address the conflicts ?

tomerweller (approved):

paulbellamy:
@chadoh Seems like the test runner is failing to launch the `soroban` cli. Is it actually being built where expected?

paulbellamy:
Aaaaaaand the top-level Cargo.toml was removed, so there's no more `make build-test-wasms`, and this is broken now.

Issue #226, opened 2023-04-10
Create a single signer account example


tomerweller:
The current custom account example is fairly advanced. I think we can benefit from a "simpler" custom account example. 

The simplest viable account I can think of is a single signer account (similar to EOA, externally owned account, on Ethereum). Should we create it as an example?

tyvdh:
I would love to see this. Account abstraction is a huge topic right now and coming up with some examples around multisig or delegated signing would quite very relevant and helpful

PR #227, opened 2023-04-11
Add a very simple account contract example.

dmkozh:
### What

Add a very simple account contract example.

This is close to minimal possible valid account contract that is aimed at demonstrating the most basic concepts.

### Why

This is in response to request for a simpler account contract example(https://github.com/stellar/soroban-examples/issues/226) - the only existing example is quite involved and aims at demonstrating all the account contract features.

### Known limitations

N/A

### DIFF
--- a/simple_account/src/test.rs
+++ b/simple_account/src/test.rs
@@ -1,0 +1,63 @@
+#![cfg(test)]
+extern crate std;
+
+use ed25519_dalek::Keypair;
+use ed25519_dalek::Signer;
+use rand::thread_rng;
+use soroban_auth::testutils::EnvAuthUtils;
+use soroban_sdk::RawVal;
+use soroban_sdk::Status;
+use soroban_sdk::{testutils::BytesN as _, vec, BytesN, Env, IntoVal};
+
+use crate::SimpleAccount;
+use crate::SimpleAccountClient;
+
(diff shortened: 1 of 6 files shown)


tomerweller (approved):
LGTM!

leighmcculloch:
Love how simple this is. 👏🏻

simple_account/src/test.rs
```diff
@@ -0,0 +1,63 @@
...
+    let payload = BytesN::random(&env);
+    // `__check_auth` can't be called directly, hence we need to use
+    // `invoke_account_contract_check_auth` testing utility that emulates being
+    // called by the Soroban host during a `require_auth` call.
+    env.invoke_account_contract_check_auth::<Status>(
+        &account_contract.contract_id,
+        &payload,
+        &vec![&env, sign(&env, &signer, &payload)],
+        &vec![&env],
+    )
```
leighmcculloch, line 49: Why do we need to use `invoke_account_contract_check_auth`? Can we test the account contract by simply calling the contract like it would be called in deployment?

dmkozh: This actually emulates the call from host - `__check_auth` function is reserved and can't be called directly (we've also recently reserved all the contract functions starting with `__` as well). The comment above tries to explain that.

PR #228, opened 2023-04-13
Added profile.release and release-with-logs to Logging example Cargo.toml

anupsdf:
Added profile.release and release-with-logs to Logging example Cargo.toml. Since the example wants users to run with these profile options. Also, added soroban-sdk version here because without it I was getting an error `Status(HostFunctionError(InputArgsInvalid))` when invoking the contract. 

After this fix, I can see the debug statement `Hello Symbol(friend)` but there is also an error after that. 
```
#0: debug: Hello Symbol(friend)
error: debug events unsupported: DebugEvent {
    msg: Some(
        "Hello {}",
    ),
    args: [
        Val(
        Symbol(friend),
    ),
    ],
}
```

### DIFF
--- a/logging/Cargo.toml
+++ b/logging/Cargo.toml
@@ -14,10 +14,10 @@ doctest = false
 testutils = ["soroban-sdk/testutils"]
 
 [dependencies]
-soroban-sdk = { workspace = true }
+soroban-sdk = "0.7.0"
 
 [dev_dependencies]
-soroban-sdk = { workspace = true, features = ["testutils"] }
+soroban-sdk = { version = "0.7.0", features = ["testutils"] }
 
 [profile.release]
 opt-level = "z"
--- a/logging/Cargo.toml
+++ b/logging/Cargo.toml
@@ -14,21 +14,7 @@ doctest = false
 testutils = ["soroban-sdk/testutils"]
 
 [dependencies]
-soroban-sdk = "0.7.0"
+soroban-sdk = { workspace = true }
 
 [dev_dependencies]
-soroban-sdk = { version = "0.7.0", features = ["testutils"] }
-
-[profile.release]
-opt-level = "z"
-overflow-checks = true
-debug = 0
-strip = "symbols"
-debug-assertions = false
-panic = "abort"
-codegen-units = 1
-lto = true
-
-[profile.release-with-logs]
-inherits = "release"
-debug-assertions = true
+soroban-sdk = { workspace = true, features = ["testutils"] }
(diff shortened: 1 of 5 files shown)


logging/Cargo.toml
```diff
@@ -14,7 +14,21 @@ doctest = false
 testutils = ["soroban-sdk/testutils"]
 
 [dependencies]
-soroban-sdk = { workspace = true }
+soroban-sdk = "0.7.0"
 
 [dev_dependencies]
-soroban-sdk = { workspace = true, features = ["testutils"] }
+soroban-sdk = { version = "0.7.0", features = ["testutils"] }
```
willemneal, line 20: I think it's worth investigating this.  It's a pain to have to change the deps for specific examples.

dmkozh: Yeah, the workspace deps have to be used everywhere. If that didn't work for you before, chances are the workspace dependency was outdated.

logging/Cargo.toml
```diff
@@ -14,7 +14,21 @@ doctest = false
...
+
+[profile.release]
```
dmkozh, line 22: Can this be defined in the root Cargo.toml? Adding the profiles to every example seems too verbose.

anupsdf: I just noticed that their is a root Cargo.toml and that it already has this profile.release. I will take this PR back to investigate why it was complaining.

leighmcculloch:
@anupsdf It sounds like you might be building the `logging` crate in isolation separate to the workspace. In the example repo all the examples build in context of the virtual workspace `Cargo.toml` file at the root.

Given that people engage with examples in the docs in isolation, then here a big connected soup, maybe we should rethink the virtual workspace in the examples repo and remove it.

While it will be a little annoying to have to update the version of the SDK in multiple files, it will create a better developer experience for developers trying out one or two examples. We should optimize for their experience.

dmkozh:
> Given that people engage with examples in the docs in isolation, then here a big connected soup, maybe we should rethink the virtual workspace in the examples repo and remove it.

Agree here, virtual workspace is pretty confusing and I'm sure I'm not the only one who initially tried to just run an example from the example directory and failed. My point was that if we do use it, then we should benefit from it. But probably a more intuitive solution is really to make every example standalone, which is of course a lot of boilerplate, but also is much friendlier to the people new to Rust.

anupsdf:
The main issue was that some examples had dependencies with other examples causing errors. I will close this one and create a new PR to fix the dependencies.

PR #229, opened 2023-04-15
Fixing the dependencies between some of the examples

anupsdf:
### What

Some examples have dependencies between them which was not specified in Cargo.toml files.

### Why

The first time the wasms were build from the root directory of `soroban-examples`, there would be errors because of the dependencies. Adding the dependencies explicitly fixes this issue. 

Command run to test the fix:
```
cargo build --target wasm32-unknown-unknown --release
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -866,6 +866,7 @@ name = "soroban-atomic-multiswap-contract"
 version = "0.0.0"
 dependencies = [
  "assert_unordered",
+ "soroban-atomic-swap-contract",
  "soroban-sdk",
 ]
 
@@ -903,6 +904,7 @@ dependencies = [
 name = "soroban-cross-contract-b-contract"
 version = "0.0.0"
 dependencies = [
+ "soroban-cross-contract-a-contract",
  "soroban-sdk",
 ]
 
@@ -1033,6 +1035,7 @@ version = "0.0.0"
 dependencies = [
  "num-integer",
  "soroban-sdk",
+ "soroban-token-contract",
 ]
 
 [[package]]
--- a/atomic_multiswap/Cargo.toml
+++ b/atomic_multiswap/Cargo.toml
@@ -11,6 +11,7 @@ crate-type = ["cdylib"]
 doctest = false
 
 [dependencies]
+soroban-atomic-swap-contract = { path = "../atomic_swap" }
 soroban-sdk = { workspace = true }
 
 [dev_dependencies]
--- a/cross_contract/contract_b/Cargo.toml
+++ b/cross_contract/contract_b/Cargo.toml
@@ -11,6 +11,7 @@ crate-type = ["cdylib"]
 doctest = false
 
 [dependencies]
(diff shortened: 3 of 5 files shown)


leighmcculloch (dismissed):
Nice find. 👍

I'm surprised we are able to reference these crates as dependencies. The reason I'm surprised is that the contract crates have their `crate-type` as only `cdylib` in their `Cargo.toml`, and don't include `lib` or `rlib`. I've seen cases where you can't import an only-`cdylib`. (We don't lib rlib as a crate-type because multi-crate-type packages don't support some optimizations and then we end up with large .wasm files.)

If it works for the examples 👍. We might find it stop working at some point if our examples change or if we add a new example that happens to do whatever causes it to no longer work. There's an unstable cargo feature ([artifact-dependencies](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies)) for supporting dependencies to other crates artifacts which will cause them to build in order without having to import them. We can switch to that if it stabilizes.

dmkozh:
> Nice find. 👍
> 
> I'm surprised we are able to reference these crates as dependencies. The reason I'm surprised is that the contract crates have their `crate-type` as only `cdylib` in their `Cargo.toml`, and don't include `lib` or `rlib`. I've seen cases where you can't import an only-`cdylib`. (We don't lib rlib as a crate-type because multi-crate-type packages don't support some optimizations and then we end up with large .wasm files.)
> 
> If it works for the examples 👍. We might find it stop working at some point if our examples change or if we add a new example that happens to do whatever causes it to no longer work. There's an unstable cargo feature ([artifact-dependencies](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies)) for supporting dependencies to other crates artifacts which will cause them to build in order without having to import them. We can switch to that if it stabilizes.

I'd be surprised if that works as expected, as the dependency contracts and user contracts typically are built with different targets (wasm for dependencies, x86 for the tests).

The artifact dependency would be great though.

leighmcculloch:
I think we should pass on the path dependency, and instruct folks to run `make build` before doing other things. Then once artifact dependencies stabilize, use it.

I think it would also be reasonable to remove the virtual workspace and include instructions for each example for how to build it. The ones that are dependent on others would include an instruction to build those other contracts first.

anupsdf:
> I think we should pass on the path dependency, and instruct folks to run `make build` before doing other things. Then once artifact dependencies stabilize, use it.
> 
> I think it would also be reasonable to remove the virtual workspace and include instructions for each example for how to build it. The ones that are dependent on others would include an instruction to build those other contracts first.

Sounds good! That way users can go to one example at a time and build them while learning. I will create a new PR to remove the virtual workspace. Closing this PR.

leighmcculloch:
> I will create a new PR to remove the virtual workspace. 

This might break the gitpod setup, but that should be trivial to fix. Fyi as this may not be obvious as it won't cause any tests to break, but opening the gitpod after opening a PR should verify if anything is broken.

PR #232, opened 2023-04-19
Removed top level Cargo.toml to have individual examples not be managed under a virtual workspace

anupsdf:
### What

Making the Soroban Examples as independent projects. 
- Removed top level Cargo.toml to have individual examples not be managed under a virtual workspace. 
- Moved all the dependencies and dev-dependencies from top level Cargo.toml to individual examples's Cargo.toml file. It was mostly soroban-sdk and some had soroban-auth. 
- Each project (example) would have its own Makefile that would be called by the root Makefile

Side note: For soroban-sdk dependency providing version alone works. So I didn't include git and rev that top level Cargo.toml had.

### Why

We are doing this because each example can be individually managed. User will go to each example directory, build them and invoke the soroban contract. For some of the examples where there is a package dependency, they would have to build the packages in the right order. In future we can use [artifact-dependency](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#artifact-dependencies) when it becomes available in stable rust channel.


leighmcculloch (commented):
Nice, this is going to be easier for developers approaching the repo from the docs.

Could we also:
- Add a `Makefile` for each project that does the basic build / test like the root one does, with the `build` step doing a build of any dependency in the examples repo as well?
- Update the `Makefile` in the root to run build / test on all the sub-folder's make files?

Makefile
```diff
@@ -1,20 +1,24 @@
+MAKEFILES = $(shell find . -maxdepth 3 -type f -name Makefile)
```
anupsdf, line 1: This is to find all the Makefile in the subdirectories.

.github/workflows/rust.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: rustup update
-    - run: cargo fmt --all --check
+    - run: make fmt
```
anupsdf, line 26: Since there is no Cargo.toml in top level we would call make fmt instead to invoke Makefile in top level that recursively calls make fmt for sub-directories.

.gitpod.yml
```diff
@@ -33,7 +33,7 @@ tasks:
       gp open increment/src/lib.rs
       gp open increment/src/test.rs
       gp open README.md
-      soroban contract invoke --id 1 --wasm target/wasm32-unknown-unknown/release/soroban_increment_contract.wasm --fn increment
+      soroban contract invoke --id 1 --wasm increment/target/wasm32-unknown-unknown/release/soroban_increment_contract.wasm --fn increment
```
anupsdf, line 36: Since there is no top level `target` directory with this PR.

Makefile
```diff
@@ -1,26 +1,42 @@
+$(MAKE)FILES = $(shell find . -maxdepth 3 -type f -name $(MAKE)file)
+SUBDIRS   = $(filter-out ./,$(dir $($(MAKE)FILES)))
+ROOT_DIR = $(shell pwd)
+
 default: build
 
 all: build test
```
anupsdf, line 7: Since `test` target internally calls `build` target, we can remove this redundant `build` from this `all` target here.

.github/workflows/rust.yml
leighmcculloch: Another option is to use `rustfmt **/*.rs` instead of `cargo fmt`. `cargo fmt` will format all rust files that are part of a project. `rustfmt` will format all rust files in any sub-directory. Remarkably this means `rustfmt` is often easier to use than the cargo equivalent.

anupsdf: Thanks for the suggestion! I have made the change to use `rustfmt **/*.rs`.

Makefile
```diff
@@ -1,26 +1,42 @@
...
+		cd $(ROOT_DIR); \
+	done
 
 watch:
-	cargo watch --clear --watch-when-idle --shell '$(MAKE)'
+	@for dir in $(SUBDIRS) ; do \
+		cd $$dir; \
+		$(MAKE) watch || break; \
+		cd $(ROOT_DIR); \
+	done
```
leighmcculloch, line 28: Watch will block on the first call and won't call others, so I don't think this will work. We can drop the watch target, it is't often used as far as I know.

timelock/Makefile
```diff
@@ -0,0 +1,20 @@
+default: build
+
+all: build test
+
+test: build
+	cargo test
+	cargo test --features testutils
+
+build:
+	cargo build --target wasm32-unknown-unknown --release -p soroban-timelock-contract
```
leighmcculloch, line 10: Don't need to specify the crate name for single crate workspaces.
```suggestion
	cargo build --target wasm32-unknown-unknown --release
```

cross_contract/contract_b/Makefile
```diff
@@ -0,0 +1,24 @@
...
+all: build test
+
+test: build
+	cargo test
+	cargo test --features testutils
+
+build:
+	@cd ../contract_a && make build || break; 
+
+	@cd $(CUR_DIR);
```
leighmcculloch, line 13: I think you can run make in an alternative directory. I think it's common practice to use `$(MAKE)` instead of `make` too, although I'm fuzzy on the why.
```suggestion
	$(MAKE) -C ../contract_a build || break; 
```

dmkozh (approved):

Makefile
```diff
@@ -1,26 +1,35 @@
...
-	cargo build --target wasm32-unknown-unknown --release -p soroban-atomic-swap-contract
-	cargo build --target wasm32-unknown-unknown --release
-	cd target/wasm32-unknown-unknown/release/ && \
-		for i in *.wasm ; do \
-			ls -l "$$i"; \
-		done
+	@for dir in $(SUBDIRS) ; do \
+		cd $$dir; \
+		$(MAKE) build || break; \
+		cd $(ROOT_DIR); \
```
leighmcculloch, line 13: You should be able to do this, which should mean you don't need the "ROOT_DIR" parameter.
```suggestion
		$(MAKE) -C $$dir build || break; \
```

anupsdf: Made the change in the new commit.

leighmcculloch (approved):
👏🏻 Looks good to me.

anupsdf:
Expanding the job [fmt](https://github.com/stellar/soroban-examples/actions/runs/4769028475/jobs/8478988812?pr=232) shows that the shell command didn't run. I will make another change to use `command` in the `run` step to see if it works.

PR #236, opened 2023-04-25
build: add Cargo.lock to each directory

chadoh:
### What

Since #232, these are no longer members of a workspace, and now running `make build` in the root adds a `target` and `Cargo.lock` to each directory.

This ignores each such `target`, but adds all the `Cargo.lock` files to version history.

### Why

Avoid very many git changes for anyone building this project locally

### Known limitations

N/A


.gitignore
```diff
@@ -1,4 +1,4 @@
-/target/
+target
 /target-tiny/
```
leighmcculloch, line 2: I believe this is no longer needed. I realize this is out-of-scope of this PR, but with loose relevance we could remove too.
```suggestion
```

leighmcculloch (approved):

PR #247, opened 2023-05-16
Make fmt target modify files

leighmcculloch:
### What
Make the `make fmt` target of projects modify files, instead of erroring and displaying a diff.

### Why
The `make fmt` target used to update unformatted files in this repository, as it does in other repositories. It appears that recently it was changed to check, error, and display a diff, which is less convenient. I think this change was so that the `fmt` GitHub Action job could call a single command to run a check. We can get the same behavior in the GitHub Action by using rustfmt directly.

<a href="https://gitpod.io/#https://github.com/stellar/soroban-examples/pull/247"></a>

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -24,8 +24,8 @@ jobs:
     - uses: actions/checkout@v3
     - run: rustup update
     - run: |
-      shopt -s globstar
-      rustfmt **/*.rs
+        shopt -s globstar
+        rustfmt **/*.rs
 
   rust-analyzer-compat:
     runs-on: ubuntu-latest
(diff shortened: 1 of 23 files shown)


anupsdf:
I had the same `file `**/*.rs` does not exist` error with rustfmt. You could try $(wildcard **/*.rs) as per this [make tutorial](https://makefiletutorial.com/).

leighmcculloch:
@anupsdf I've updated the PR now and it passes. I'm avoiding using the makefile target, because I changed the makefile target not to error and display a diff, but instead fix the formatting.

anupsdf (approved):
LGTM

tomerweller (approved):

PR #248, opened 2023-05-17
Run each example in isolation in CI

leighmcculloch:
### What
Run each example in isolation in CI

### Why
Somebody previously suggested this (@MonsieurNicolas maybe?), although I don't remember why we didn't do it then.

I'd like to do it now so it is clearer what is failing when we're updating SDKs, or accepting contributions.

Now that we have each example isolated as its own workspace, running make test fails early on the first example that hits an error. It'd be helpful when looking at PR CI if the failures were across the entire set of examples so we understood how far away a PR is from being ready.

Running in isolation may not be faster as a single run uses a lot of instances for a short period of time. It shouldn't be prohibitively long though.

<a href="https://gitpod.io/#https://github.com/stellar/soroban-examples/pull/248"></a>

### DIFF
--- a/custom_types/src/lib.rs
+++ b/custom_types/src/lib.rs
@@ -33,7 +33,10 @@ impl IncrementContract {
     pub fn get_state(env: Env) -> State {
         env.storage()
             .get(&STATE)
-            .unwrap_or_else(|| Ok(State::default())) // If no value set, assume 0.
+            .unwrap_or(Ok(State {
+                count: 0,
+                last_incr: 0,
+            })) // If no value set, assume 0.
             .unwrap() // Panic if the value of COUNTER is not a State.
     }
 }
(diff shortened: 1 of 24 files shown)


custom_types/src/lib.rs
```diff
@@ -33,7 +33,10 @@ impl IncrementContract {
     pub fn get_state(env: Env) -> State {
         env.storage()
             .get(&STATE)
-            .unwrap_or_else(|| Ok(State::default())) // If no value set, assume 0.
+            .unwrap_or(Ok(State {
+                count: 0,
+                last_incr: 0,
+            })) // If no value set, assume 0.
```
leighmcculloch, line 39: I'm surprised I need to fix this in this PR as I thought I was just changing how tests ran in CI, but I think I might have in the process plugged a bug in the CI run that wasn't running all the builds / tests, because the custom_types crate is failing to build after I changed the CI workflow.

sisuresh (approved):

PR #275, opened 2023-09-01
Use sdk token interface instead of own interface in example

leighmcculloch:
### What
Use sdk token interface instead of own interface in token example.

### Why
The SDK provides a standard token interface in hope that token contracts will export a consistent interface promoting interoperability, in a similar way that the Ethereum ecosystem has rallied around ERC20 and other standards.

The example should lean on that interface to encourage others to, to continue to maintain consistency.

### DIFF
--- a/token/src/contract.rs
+++ b/token/src/contract.rs
@@ -7,44 +7,9 @@ use crate::balance::{read_balance, receive_balance, spend_balance};
 use crate::event;
 use crate::metadata::{read_decimal, read_name, read_symbol, write_metadata};
 use crate::storage_types::INSTANCE_BUMP_AMOUNT;
+use soroban_sdk::token::{self, Interface as _};
 use soroban_sdk::{contract, contractimpl, Address, Env, String};
 use soroban_token_sdk::TokenMetadata;
-
-pub trait TokenTrait {
-    fn initialize(e: Env, admin: Address, decimal: u32, name: String, symbol: String);
-
-    fn allowance(e: Env, from: Address, spender: Address) -> i128;
-
-    fn approve(e: Env, from: Address, spender: Address, amount: i128, expiration_ledger: u32);
-
-    fn balance(e: Env, id: Address) -> i128;
-
-    fn spendable_balance(e: Env, id: Address) -> i128;
-
-    fn authorized(e: Env, id: Address) -> bool;
-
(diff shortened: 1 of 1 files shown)


token/src/contract.rs
```diff
@@ -76,6 +41,51 @@ impl TokenTrait for Token {
...
+
+        e.storage().instance().bump(INSTANCE_BUMP_AMOUNT);
+
+        write_administrator(&e, &new_admin);
+        event::set_admin(&e, admin, new_admin);
+    }
+}
+
+#[contractimpl]
+impl token::Interface for Token {
```
leighmcculloch, line 88: It's unfortunate that we named the interface type simply `Interface`. Technically doing so adheres to some Rust best practice not to repeat the module name in the type. However, in importers we end up with simple `Interface` which does a poor job of communicating intent. Especially if there are multiple interfaces being implemented. It would have been better to name it `TokenInterface` I think.

So for the sake of readability importing it with the `token::` qualifier.

```rust
impl token::Interface for Token {
```

Note that this means that the import line is a little bonkers. It imports `token::self` so that `token` can be referenced. And it has to separately import `Interface` because implemented traits has to be imported into the scope to be used, otherwise the compiler will error on their use. The `as _` is so that `Interface` is imported into the scope without being given a name that must be used, since it is never referenced and would considered an unused import.

```rust
use soroban_sdk::token::{self, Interface as _};
```

token/src/contract.rs
```diff
@@ -155,48 +165,6 @@ impl TokenTrait for Token {
...
-
-    fn set_admin(e: Env, new_admin: Address) {
-        let admin = read_administrator(&e);
-        admin.require_auth();
-
-        e.storage().instance().bump(INSTANCE_BUMP_AMOUNT);
-
-        write_administrator(&e, &new_admin);
-        event::set_admin(&e, admin, new_admin);
-    }
```
leighmcculloch, line 198: A nice side-effect of this change is that the example contract is clearer about which functions it exposes are part of the token interface and which are its own functions that provide functionality that isn't described by the standard token interface.

graydon (approved):

leighmcculloch:
I was going to merge this to a branch for the next release. But the changes in this PR doesn't meaningful manifest themselves into the docs for this example. Because the example is so large only a small subset of its code is actually embedded in the docs. So I think this change won't present a significant inconsistency.

PR #276, opened 2023-09-01
Simplify token example removing authorization and clawback

leighmcculloch:
### What
Remove authoriation and clawback from the token example.

### Why
Clawback is not a feature of the standard token. It's a little odd in a lightweight token example to include functionality that isn't part of the standard token interface. It is part of the Stellar Asset contract, but other tokens do not need to provide clawback functionality. So it muddies the water on what is or isn't part of the standard token interface to provide it.

Likewise, I don't think authorization should be part of the standard token interface and I've opened a PR proposing its removal in https://github.com/stellar/rs-soroban-sdk/pull/1079. Authorization logic could be so varied, and many contracts are unlikely to have it. For the moment I've simplified the authorization logic in the example contract to simply say that all addresses are authorized.

As a side-effect these changes simplify the token example, which is helpful. The standard token has enough functions as it is for folks to wrap their minds arounds. Everything we can do to help them step into building their own token with reduced complexity and concerns lowers the barrier.

### DIFF
--- a/token/src/balance.rs
+++ b/token/src/balance.rs
@@ -19,33 +19,13 @@ fn write_balance(e: &Env, addr: Address, amount: i128) {
 
 pub fn receive_balance(e: &Env, addr: Address, amount: i128) {
     let balance = read_balance(e, addr.clone());
-    if !is_authorized(e, addr.clone()) {
-        panic!("can't receive when deauthorized");
-    }
     write_balance(e, addr, balance + amount);
 }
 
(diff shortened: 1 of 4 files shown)


leighmcculloch:
Similar to #275, I was going to merge this to a branch for the next release. But the changes in this PR doesn't meaningful manifest themselves into the docs for this example. Because the example is so large only a small subset of its code is actually embedded in the docs. So I think this change won't present a significant inconsistency.

sisuresh (approved):

PR #298, opened 2024-01-19
Add example with ecdsa secp256r1 (prime256r1 / p256) verification

leighmcculloch:
### What
Add example with ecdsa secp256r1 (prime256r1 / p256) verification.

### Why
To demonstrate how to use crypto libraries to do other forms of crypto in the guest side.

Thanks to @graydon for helping get me started on this.


secp256r1/src/test.rs
```diff
@@ -0,0 +1,38 @@
...
+#[test]
+fn test_wasm() {
+    let env = Env::default();
+    let contract_id = env.register_contract_wasm(None, WASM);
+    test(&env, &contract_id);
+}
+
+fn test(env: &Env, contract_id: &Address) {
+    // Set an unlimited budget because it'll fail with the default budget.
+    env.budget().reset_unlimited();
```
leighmcculloch, line 24: When running the `test_wasm` it fails if the budget is the default budget, even if I reset the budget just before calling verify. I only seem to be able to get it to pass if I first reset to an unlimited budget. cc @graydon @tomerweller

leighmcculloch:
I may reopen again in the future with a copy of https://github.com/leighmcculloch/soroban-webauthn.

PR #301, opened 2024-02-27
Add eth abi example

leighmcculloch:
### What
Add eth abi example.

### Why
Interoperability with contracts that use the ETH ABI for encoding messages and data is likely given its adoption in other chains.


tomerweller (approved):
lgtm

PR #303, opened 2024-04-05
Update custom account examples to use the SDK interface

leighmcculloch:
### What
Update custom account examples to use the SDK interface.

### Why
The interface is there to help make sure that implementations match the interface that the Soroban Env expects. The examples should encourage use of them, even though they aren't required.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -16,7 +16,7 @@ struct AccountContract;
 
 #[contracttype]
 #[derive(Clone)]
-pub struct Signature {
+pub struct AccSignature {
     pub public_key: BytesN<32>,
     pub signature: BytesN<64>,
 }
@@ -73,7 +73,7 @@ impl AccountContract {
 
 #[contractimpl]
 impl CustomAccountInterface for AccountContract {
-    type Signature = Vec<Signature>;
+    type Signature = Vec<AccSignature>;
     type Error = AccError;
 
     // This is the 'entry point' of the account contract and every account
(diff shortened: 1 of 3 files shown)


account/src/lib.rs
```diff
@@ -68,6 +69,12 @@ impl AccountContract {
             .instance()
             .set(&DataKey::SpendLimit(token), &limit);
     }
+}
+
+#[contractimpl]
+impl CustomAccountInterface for AccountContract {
+    type Signature = Vec<Signature>;
```
dmkozh, line 76: Using `Signature` as both type and type alias is really confusing (I've been scratching my head while looking at this). Can we just call the struct `AccSignature` or something else, different from `Signature`? Also maybe use the alias in the implementation as well, though that's confusing in a different way.

leighmcculloch: It's really common in rust, at least from what I've seen, to shadow and reuse the same names for things.

In this case the generic associated type can coexist with the concrete type because the generic associated type must always be qualified with a Self or the parent types name. The qualification is what to look for when identifying which Signature it refers. Also it's worth noting that even if you mix them up it makes no conceptual difference. The Signature is still the Signature, so conveniently by using the same name we reduce the number of names one must reason about. i.e. the distinction doesn't really matter so if you ignore it the cognitive load is reduced.

I can change it before merging, but I think you'll find this pattern is common and will show up in other rust code.

leighmcculloch: Correction, I forgot that in this case the associated type is actually a Vec of signature makes this more confusing and it is not true about the cognitive load. I'll update it.

dmkozh: Exactly, when you look at `Vec<Signature>` it's actually not immediately apparent whether it's actually a `Vec<Vec<Signature>>` or ``Vec<Signature>` (at least that's how I read it first, sorry for not mentioning that in the initial comment).

leighmcculloch: Yup 👍 Good call.

dmkozh (approved):

PR #304, opened 2024-04-05
Update Cargo.toml file on all contracts.

fnando:
### What

Simplify all Cargo.toml files, matching what's generated by `soroban contract init`.

### Why

`soroban contract init --with-example <name>` uses this repo as the source. By making these changes, we are able to copy examples that are closer to the actual generated contract.

### Known limitations

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@ target/
 /.sccache/
 .DS_Store
 .soroban
-errors/test_snapshots/
+test_snapshots
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target/
 /.sccache/
 .DS_Store
 .soroban
+errors/test_snapshots/
--- a/account/Cargo.toml
+++ b/account/Cargo.toml
@@ -1,10 +1,7 @@
 [package]
 name = "soroban-account-contract"
 version = "0.0.0"
-authors = ["Stellar Development Foundation <info@stellar.org>"]
-license = "Apache-2.0"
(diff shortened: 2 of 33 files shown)


account/Cargo.toml
```diff
@@ -1,10 +1,7 @@
 [package]
 name = "soroban-account-contract"
 version = "0.0.0"
-authors = ["Stellar Development Foundation <info@stellar.org>"]
-license = "Apache-2.0"
```
leighmcculloch, line 5: The examples are authored by SDF and licensed as Apache 2 so I think we should probably keep them specified as such? Or if we drop the author at least keep the license. I guess we have the license in the root so it doesn't really matter what's in the toml. 

Lgtm 👍

leighmcculloch (approved):

.gitignore
```diff
@@ -2,3 +2,4 @@ target/
 /.sccache/
 .DS_Store
 .soroban
+errors/test_snapshots/
```
kalepail, line 5: Should we ignore all test snapshot directories?

account/Cargo.toml
kalepail: I vote toss. If we're going to use these as templates we should simplify the tomls to help downstream folks not accidentally be adding authors and licenses they don't intend to

kalepail (approved):

.gitignore
leighmcculloch: Yup with `**/test_snapshots/`.

PR #305, opened 2024-04-07
Token with supply

icolomina:
A contract example which is similar to the token example but the token requires a supply:
- initialize function requires supply
- mint decrements supply
- Adds a supply function to query the current supply
- burn and burn_from functions increments supply with the burned amount. Those tokens are out or the circulation after burned but they can be minted in the future.

We agreed in [this pull request](https://github.com/stellar/soroban-examples/pull/302) that it would be better to create another example instead of modifying the existing token example with the supply changes since supply may be not ideal to include on most tokens on Soroban.


leighmcculloch:
Asking a question about this here:
- https://github.com/stellar/soroban-examples/pull/302#discussion_r1555217041

PR #307, opened 2024-04-16
Prevent contracts from being re-initialized

matiassequeira:
What
Prevent the init function front being executed after the contract has already been initialized.

Why
Incorporating a security feature to prevent adversaries from substituting the admin by invoking the init function, enabling them to upgrade the contract to a malicious version. This fix prevents anyone copying the upgrade functionality from replicating this vulnerability in their contracts.

Known limitations
N/A

### DIFF
--- a/upgradeable_contract/new_contract/src/lib.rs
+++ b/upgradeable_contract/new_contract/src/lib.rs
@@ -22,7 +22,7 @@ pub struct UpgradeableContract;
 impl UpgradeableContract {
     pub fn init(e: Env, admin: Address) -> Result<(), Error> {
         if e.storage().instance().has(&DataKey::Admin) {
-            return Err(Error::AlreadyInitialized)
+            return Err(Error::AlreadyInitialized);
         }
         e.storage().instance().set(&DataKey::Admin, &admin);
         Ok(())
--- a/upgradeable_contract/new_contract/src/lib.rs
+++ b/upgradeable_contract/new_contract/src/lib.rs
@@ -1,6 +1,6 @@
 #![no_std]
 
-use soroban_sdk::{contract, contractimpl, contracttype, Address, BytesN, Env};
+use soroban_sdk::{contract, contracterror, contractimpl, contracttype, Address, BytesN, Env};
 
 #[contracttype]
 #[derive(Clone)]
@@ -8,13 +8,24 @@ enum DataKey {
     Admin,
 }
 
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum Error {
+    AlreadyInitialized = 1,
+}
+
 #[contract]
 pub struct UpgradeableContract;
 
 #[contractimpl]
 impl UpgradeableContract {
-    pub fn init(e: Env, admin: Address) {
+    pub fn init(e: Env, admin: Address) -> Result<(), Error> {
+        if e.storage().instance().has(&DataKey::Admin) {
+            return Err(Error::AlreadyInitialized)
+        }
         e.storage().instance().set(&DataKey::Admin, &admin);
+        Ok(())
     }
 
     pub fn version() -> u32 {
--- a/upgradeable_contract/old_contract/src/lib.rs
+++ b/upgradeable_contract/old_contract/src/lib.rs
@@ -22,7 +22,7 @@ pub struct UpgradeableContract;
 impl UpgradeableContract {
     pub fn init(e: Env, admin: Address) -> Result<(), Error> {
         if e.storage().instance().has(&DataKey::Admin) {
-            return Err(Error::AlreadyInitialized)
+            return Err(Error::AlreadyInitialized);
         }
         e.storage().instance().set(&DataKey::Admin, &admin);
         Ok(())
--- a/upgradeable_contract/old_contract/src/lib.rs
+++ b/upgradeable_contract/old_contract/src/lib.rs
@@ -1,6 +1,6 @@
 #![no_std]
 
-use soroban_sdk::{contract, contractimpl, contracttype, Address, BytesN, Env};
+use soroban_sdk::{contract, contracterror, contractimpl, contracttype, Address, BytesN, Env};
 
 #[contracttype]
 #[derive(Clone)]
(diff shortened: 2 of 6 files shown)


upgradeable_contract/old_contract/src/test.rs
```diff
@@ -44,3 +44,20 @@ fn test() {
     let client = new_contract::Client::new(&env, &contract_id);
     assert_eq!(1010101, client.new_v2_fn());
 }
+
+#[test]
+#[should_panic(expected = "HostError: Error(Contract, #1)")]
```
dmkozh, line 49: Could we use `try_init` and an assertion on the returned error instead of panic? The current setup is kind of brittle, e.g. it allows first `init` call to panic.

upgradeable_contract/old_contract/src/lib.rs
```diff
@@ -1,20 +1,31 @@
...
+
 #[contract]
 pub struct UpgradeableContract;
 
 #[contractimpl]
 impl UpgradeableContract {
-    pub fn init(e: Env, admin: Address) {
+    pub fn init(e: Env, admin: Address) -> Result<(), Error> {
+        if e.storage().instance().has(&DataKey::Admin) {
+            return Err(Error::AlreadyInitialized)
```
dmkozh, line 25: nit: Please add semicolon

upgradeable_contract/new_contract/src/lib.rs
```diff
@@ -1,20 +1,31 @@
...
 #[derive(Clone)]
 enum DataKey {
     Admin,
 }
 
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum Error {
+    AlreadyInitialized = 1,
```
dmkozh, line 15: nit: Please add semicolon

matiassequeira: You mean, semicolon after the enum closing bracket?

upgradeable_contract/old_contract/src/test.rs
matiassequeira: Sure. Note however I followed the examples with `try_...` and I'm getting the following compilation error:

```
error[E0308]: mismatched types
  --> src/test.rs:61:41
   |
61 |     assert_eq!(client.try_init(&admin), Err(Ok(Error::AlreadyInitialized)),);
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<Result<(), ConversionError>, Result<..., ...>>`, found `Result<_, Result<Error, _>>`
```

The test looks like this now:
```
fn test_cannot_re_init() {
    let env = Env::default();
    env.mock_all_auths();

    // Note that we use register_contract_wasm instead of register_contract
    // because the old contracts WASM is expected to exist in storage.
    let contract_id = env.register_contract_wasm(None, old_contract::WASM);

    let client = old_contract::Client::new(&env, &contract_id);
    let admin = Address::generate(&env);
    client.init(&admin);
    assert_eq!(client.try_init(&admin), Err(Ok(Error::AlreadyInitialized)));
}
```

And I've imported the error with `use super::*;`

Any idea of what could be wrong? It sounds like a problem with the Err(Ok()) wrappers?

dmkozh: The expected result looks correct to me (the inner result for the error case can be error itself in case if your contract panics without returning your error). Could you maybe commit this change, so that we could patch and debug?

upgradeable_contract/old_contract/src/test.rs
```diff
@@ -44,3 +45,18 @@ fn test() {
...
+    env.mock_all_auths();
+
+    // Note that we use register_contract_wasm instead of register_contract
+    // because the old contracts WASM is expected to exist in storage.
+    let contract_id = env.register_contract_wasm(None, old_contract::WASM);
+
+    let client = old_contract::Client::new(&env, &contract_id);
+    let admin = Address::generate(&env);
+    client.init(&admin);
+    assert_eq!(client.try_init(&admin), Err(Ok(Error::AlreadyInitialized)));
```
matiassequeira, line 61: @dmkozh PTAL. The error I got is 

```
❯ cargo test
   Compiling soroban-upgradeable-contract-old-contract v0.0.0 (/Users/matseq/Documents/soroban-playground/2fork-soroban-examples/upgradeable_contract/old_contract)
error[E0308]: mismatched types
  --> src/test.rs:61:41
   |
61 |     assert_eq!(client.try_init(&admin), Err(Ok(Error::AlreadyInitialized)));
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<Result<(), ConversionError>, ...>`, found `Result<_, Result<Error, _>>`
   |
   = note: `Error` and `soroban_sdk::Error` have similar names, but are actually distinct types
```

dmkozh: It seems like you just need to remove `use super::*;` (that should be removed in any case) and use `crate::Error` instead.

matiassequeira: Added `use crate::Error;` and removed `use super::*;`, but still getting a similar error 

```
error[E0308]: mismatched types
  --> src/test.rs:61:41
   |
61 |     assert_eq!(client.try_init(&admin), Err(Ok(Error::AlreadyInitialized)));
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<Result<(), ...>, ...>`, found `Result<_, Result<Error, _>>`
```

dmkozh: Sorry, that's indeed pretty confusing. IIUC the client generated from Wasm doesn't use your contract error and returns a generic SDK error instead, while in the other examples the client is generated from the Rust struct directly and has a correct type. So the conversion is necessary to assert on the final error. So something like this works (I realize it's uglier, but I'm not sure we can simplify much more):
```rust
    // `try_init` is expected to return an error. Since client is generated from Wasm,
    // this is a generic SDK error.
    let err: soroban_sdk::Error = client.try_init(&admin).err().unwrap().unwrap();
    // Convert the SDK error to the contract error.
    let contract_err: Error = err.try_into().unwrap();
    // Make sure contract error has the expected value.
    assert_eq!(contract_err, Error::AlreadyInitialized);
```

matiassequeira: Got it! Took quite a few unwraps 😄 . I decided to leave your comments in the code to serve as an explanatory example of how to do it this way. It should be good to go now. Thanks!

dmkozh: Thanks for bearing with me!

matiassequeira:
Hi @dmkozh, all the checks pass now. It requires your approval again, thanks.

PR #318, opened 2024-07-17
Update to v21

leighmcculloch:
### What
Update to v21

### Why
The examples are still on SDK v20.

### DIFF
--- a/account/src/lib.rs
+++ b/account/src/lib.rs
@@ -161,7 +161,7 @@ fn authenticate(
         }
         env.crypto().ed25519_verify(
             &signature.public_key,
-            &signature_payload.to_bytes().into(),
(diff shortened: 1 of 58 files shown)


account/src/lib.rs
```diff
@@ -160,7 +161,7 @@ fn authenticate(
         }
         env.crypto().ed25519_verify(
             &signature.public_key,
-            &signature_payload.clone().into(),
+            &signature_payload.to_bytes().into(),
```
leighmcculloch, line 164: This change is because `Hash<>`, that replaced `BytesN<>` for this use case, does not implement `Clone`. Fixing that in:
- https://github.com/stellar/rs-soroban-sdk/pull/1302

fnando (approved):

PR #325, opened 2024-09-13
Add an example deploy script to hello world example

leighmcculloch:
### What
Add an example deploy script to hello world example.

### Why
I wanted to see how well the js-stellar-sdk as it is could support scripting, so I wrote a deploy script for the hello world example.

This won't be committed. The script is too complex. We'd need some simpler utilities in the js-stellar-sdk before this was really viable.

I think this is something the js-stellar-sdk could support though?

Thoughts @janewang @fnando @chadoh @willemneal @Ifropc @Shaptic ?

_Yeah, I used Deno when hacking this up, it's what I'm most familiar with. I assume we'd use some sort of npm thing._

### DIFF
--- a/hello_world/deno.jsonc
+++ b/hello_world/deno.jsonc
@@ -1,0 +1,12 @@
+{
+  "tasks": {
+    "build": "stellar contract build && ls -l target/wasm32-unknown-unknown/release/*.wasm",
+    "test": "cargo test",
+    "deploy": "./tasks/deploy.ts"
+  },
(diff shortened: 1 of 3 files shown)


Ifropc:
> The script is too complex

Yes, I agree, this is way too much code for something fundamental. We can use cli for now, so I don't think it's a huge issue at the moment, but it adding (an easy to use) deployment support in the SDK would definitely made it more convenient to write tests for external networks

PR #326, opened 2024-09-17
add option object vec

BlaineHeffron:
### What

Adds a test case for a js-stellar-sdk issue

### Why

When retrieving an Option<Vec<Object>> in the sdk you get the following error: TypeError: Type [object Object] was not vec, but [object Object] is

### Known limitations

We need to create a separate repo for test vectors but I put this here for now for the sdk bug report.

### DIFF
--- a/other_custom_types/src/lib.rs
+++ b/other_custom_types/src/lib.rs
@@ -199,4 +199,10 @@ impl CustomTypesContract {
     pub fn tuple_strukt(_env: Env, tuple_strukt: TupleStruct) -> TupleStruct {
         tuple_strukt
     }
+
+    pub fn get_object_vec_option(env: Env) -> Option<Vec<Test>> {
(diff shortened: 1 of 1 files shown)


other_custom_types/src/lib.rs
```diff
@@ -199,4 +199,18 @@ impl CustomTypesContract {
     pub fn tuple_strukt(_env: Env, tuple_strukt: TupleStruct) -> TupleStruct {
         tuple_strukt
     }
+
+    pub fn get_object_vec_option(env: Env) -> Option<Vec<Test>> {
```
leighmcculloch, line 203: For consistency with other functions in this contract, we should drop the `get_` prefix and use the same names as other functions here.
```suggestion
    pub fn option_vec_strukt(env: Env) -> Option<Vec<Test>> {
```

PR #327, opened 2024-09-23
Update Gitpod Environment and Soroban SDK

overcat:
### What

- Software used in Gitpod has been updated
- GitPod now uses the testnet Soroban RPC by default
- Soroban SDK updated to version 21.7.1

### Why

Make the GitPod environment available again. This is very helpful for beginners, as they do not need to configure the environment locally.

### Known limitations

N/A

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  MINIMUM_RUST_VERSION: 1.81.0
+  MINIMUM_RUST_VERSION: 1.74.0
 
 jobs:
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -77,9 +77,9 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - name: Install cargo binstall
-      run: curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
-    - run: cargo binstall -y stellar-cli
+    - uses: stellar/stellar-cli@3fd8cec40ab1a77b86b22e086607eebd3a80b7fb
+      with:
+        version: 21.5.0
     - run: make test
       env:
         CARGO_BUILD_TARGET: ${{ matrix.sys.target }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  MINIMUM_RUST_VERSION: 1.74.0
+  MINIMUM_RUST_VERSION: 1.81.0
 
 jobs:
 
@@ -77,10 +77,8 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - uses: stellar/binaries@v16
-      with:
-        name: soroban-cli
-        version: 0.8.7
+    - run: cargo install --locked cargo-binstall
+    - run: cargo binstall -y stellar-cli
     - run: make test
       env:
         CARGO_BUILD_TARGET: ${{ matrix.sys.target }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -77,7 +77,7 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - uses: stellar/stellar-cli@3fd8cec40ab1a77b86b22e086607eebd3a80b7fb
+    - uses: stellar/stellar-cli@d538ea9a1b5bbf1847a987ccf316d34e55465313
       with:
         version: 21.5.0
     - run: make test
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -77,7 +77,8 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - run: cargo install --locked cargo-binstall
+    - name: Install cargo binstall
+      run: curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
     - run: cargo binstall -y stellar-cli
     - run: make test
       env:
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -6,10 +6,10 @@ RUN rustup self uninstall -y
 RUN rm -rf .rustup
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y
 
-RUN rustup install stable
-RUN rustup target add --toolchain stable wasm32-unknown-unknown
-RUN rustup component add --toolchain stable rust-src
-RUN rustup default stable
+RUN rustup install 1.81
+RUN rustup target add --toolchain 1.81 wasm32-unknown-unknown
+RUN rustup component add --toolchain 1.81 rust-src
+RUN rustup default 1.81
 
 RUN sudo apt-get update && sudo apt-get install -y binaryen
 
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,16 +1,4 @@
-FROM gitpod/workspace-full:2023-01-16-03-31-28
-
-RUN mkdir -p ~/.local/bin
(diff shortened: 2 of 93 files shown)


.gitpod.Dockerfile
```diff
@@ -1,30 +1,26 @@
...
 # https://github.com/gitpod-io/workspace-images/issues/933#issuecomment-1272616892
 RUN rustup self uninstall -y
 RUN rm -rf .rustup
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y
 
-RUN rustup install 1.69
-RUN rustup target add --toolchain 1.69 wasm32-unknown-unknown
-RUN rustup component add --toolchain 1.69 rust-src
-RUN rustup default 1.69
+RUN rustup install stable
```
overcat, line 9: There can be a more concise way of writing this, but I think keeping the current state is also acceptable.

.gitpod.yml
```diff
@@ -31,25 +21,13 @@ tasks:
...
       gp open increment/src/test.rs
       gp open README.md
-      soroban contract invoke --id 1 --wasm increment/target/wasm32-unknown-unknown/release/soroban_increment_contract.wasm -- increment
+      stellar contract inspect --wasm increment/target/wasm32-unknown-unknown/release/soroban_increment_contract.wasm
 vscode:
   extensions:
     - rust-lang.rust-analyzer
     - vadimcn.vscode-lldb
 
-github:
```
overcat, line 40: The configuration has been deprecated.
https://www.gitpod.io/changelog/activity-based-prebuilds

overcat:
Hi @sisuresh @leighmcculloch, do you have time to take a look and see if this PR can be merged?

.github/workflows/rust.yml
```diff
@@ -77,10 +77,9 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - uses: stellar/binaries@v16
-      with:
-        name: soroban-cli
-        version: 0.8.7
+    - name: Install cargo binstall
+      run: curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+    - run: cargo binstall -y stellar-cli
```
leighmcculloch, line 82:
```suggestion
    - uses: stellar/stellar-cli@3fd8cec40ab1a77b86b22e086607eebd3a80b7fb
      with:
        version: 21.5.0
```

.gitpod.Dockerfile
```diff
@@ -1,30 +1,26 @@
...
 
 RUN sudo apt-get update && sudo apt-get install -y binaryen
 
 # Enable sparse registry support, which will cause cargo to download only what
 # it needs from crates.io, rather than the entire registry.
 ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+
+# Install cargo binstall
+RUN curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+RUN cargo binstall -y stellar-cli cargo-watch sccache
```
leighmcculloch, line 22: Thanks for reordering the tools to install after the other things, that makes more sense. 👍🏻 

Could we keep the existing curl binaries install process? We shouldn't need to change the process, just update the versions I would expect?

.gitpod.Dockerfile
leighmcculloch: 👍🏻

.gitpod.yml
```diff
@@ -2,24 +2,14 @@ image:
...
-      stellar/quickstart:soroban-dev \
-      --futurenet \
-      --enable-soroban-rpc
-  # This "CLI - Futurenet" task opens a terminal for you to interact with the
-  # futurenet network.
-  - name: CLI - Futurenet
-    # We specify some required environment variables for use on the futurenet.
+  # This "CLI - Testnet" task opens a terminal for you to interact with the
+  # testnet network.
+  - name: CLI - Testnet
```
leighmcculloch, line 7: @kalepail What's your opinion on connection to testnet vs running a local quickstart. A local quickstart is faster to iterate on because it has 1 second ledgers, but deploy to testnet is probably also the most direct route to feeling the rush of success of deploying to a real network.

workspace/Cargo.toml
```diff
@@ -1,17 +1,13 @@
 [workspace]
 resolver = "2"
 
-members = [
-    "contract_a_interface",
-    "contract_a",
-    "contract_b",
-]
+members = ["contract_a_interface", "contract_a", "contract_b"]
```
leighmcculloch, line 4:
```suggestion
members = [
    "contract_a_interface",
    "contract_a",
    "contract_b",
]
```

leighmcculloch (commented):
Thanks for churning through the issues to get this updated.

I've made a few requests inline, otherwise this looks great.

.gitpod.Dockerfile
overcat: Hi @leighmcculloch, I don't oppose using the existing method, but I think using `cargo binstall` would make everything look cleaner. In fact, `cargo binstall` also downloads binary files from the GitHub release page. If we need to specify a version, we can do that here too. However, in this scenario, I believe providing users with the latest version should be sufficient, and this way we won't have to update this file every time.

.github/workflows/rust.yml
```diff
@@ -77,10 +77,9 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - uses: stellar/binaries@v16
+    - uses: stellar/stellar-cli@3fd8cec40ab1a77b86b22e086607eebd3a80b7fb
```
leighmcculloch, line 80:
```suggestion
    - uses: stellar/stellar-cli@d538ea9a1b5bbf1847a987ccf316d34e55465313
```

.gitpod.Dockerfile
```diff
@@ -1,30 +1,26 @@
...
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y
 
-RUN rustup install 1.69
-RUN rustup target add --toolchain 1.69 wasm32-unknown-unknown
-RUN rustup component add --toolchain 1.69 rust-src
-RUN rustup default 1.69
+RUN rustup install stable
+RUN rustup target add --toolchain stable wasm32-unknown-unknown
+RUN rustup component add --toolchain stable rust-src
+RUN rustup default stable
```
leighmcculloch, line 12: We need to stay below 1.82 until this issue is resolved:
- https://github.com/stellar/rs-soroban-env/issues/1469
```suggestion
RUN rustup install 1.81
RUN rustup target add --toolchain 1.81 wasm32-unknown-unknown
RUN rustup component add --toolchain 1.81 rust-src
RUN rustup default 1.81
```

.github/workflows/rust.yml
```diff
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  MINIMUM_RUST_VERSION: 1.74.0
+  MINIMUM_RUST_VERSION: 1.81.0
```
leighmcculloch, line 13: The minimum rust version is still 1.74 in the version of the sdk in use in this change, so I don't think we need to change this. The minimum rust version named here, checks that the examples all still build with the minimum rust version of the SDK.
```suggestion
  MINIMUM_RUST_VERSION: 1.74.0
```

.gitpod.Dockerfile
leighmcculloch: I have not been a fan of employing cargo bininstall because it lacks supply chain security, doesnt use github attestations, compared to other tools in the space like homebrew, but to your point, it just also pulls from gh releases, which is no different to our existing approach and so doesn't actually have an advantage in regards to these things because it also doesn't do anything for supply chain security.

Let's run with this, but I'd like to circle back on it to making this more secure.

.gitpod.yml
leighmcculloch: Going to move ahead with this. Because I realise this wasn't running a local quickstart in isolation, but using futurenet, which I missed, and yeah folks should use testnet now.

.gitpod.Dockerfile
overcat: I understand your concerns, and I will modify it to the previous way. bininstall has a [signing mechanism](https://github.com/cargo-bins/cargo-binstall/blob/main/SIGNING.md), but it seems not to be widely used.

leighmcculloch: Let's leave it as you have it, as there's little advantage over it vs using bininstall, both suffer from the same problems. I'll merge this PR shortly.

leighmcculloch (approved):

.gitpod.Dockerfile
leighmcculloch: I'll have a little think and might change it in a follow up.

leighmcculloch:
@overcat Thanks for pushing us to get this fixed and updated and doing the hard work to figure out all the problems and make this happen 💪🏻.

PR #331, opened 2024-10-14
Update stellar-cli install in CI

leighmcculloch:
### What

Replace the stellar/binaries install method for the stellar-cli with the stellar-cli action.

### Why

To remove the stellar/binaries dependency that is no longer being maintained for the stellar-cli.


leighmcculloch:
Deferring to https://github.com/stellar/soroban-examples/pull/327 and moving this change into that PR.

PR #336, opened 2024-10-16
Add ci to check gitpod docker builds

leighmcculloch:
### What
Add ci to check gitpod docker builds

### Why
It recently broke and we should check the image builds, and that's a simple thing we can do to improve it's stability.


leighmcculloch:
This build will pass after #335 is merged.

fnando (approved):

PR #344, opened 2024-11-27
feat: structure as cargo workspace

### LINKED ISSUE #328: Structure as workspace
### What problem does your feature solve?

We have started encouraging people to set up projects using workspaces. For example, `stellar contract init` allows initializing a project with one of these `soroban-examples` examples, using the `--with-example` option. But the CLI then needs to tweak the `Cargo.toml` files of the examples to work as part of a workspace.

### What would you like to see?

If this were already structured as a workspace, we wouldn't need any special handling within commands like `stellar contract init`. In fact, we wouldn't really need `stellar contract init` at all. A simple [degit](https://github.com/Rich-Harris/degit) would work. For example, if you only wanted one example contract, you could do something like this:

```
degit stellar/soroban-examples/increment contracts/increment
```

This would then allow us to circle back and simplify `stellar contract init` as well, perhaps using [degit-rs](https://crates.io/crates/degit). This fits in with a larger goal of simplifying `contract init`:

- https://github.com/stellar/stellar-cli/issues/1586

### How to implement

It's really nice that the examples here are all in the top level of the repo. To keep that, we could structure this the way that we structured [stellar/soroban-test-examples](https://github.com/stellar/soroban-test-examples):

```toml
[workspace]
members = [
  "./*",
]
exclude = [
  ".git",
  "target",
]
```

chadoh:
Fixes #328

Undoes #232

Forked from #330
- [ ] TODO: Rebase onto `main` once it's merged

Add top-level `Cargo.toml`, with all project folders as part of this 
workspace.

We now encourage everyone to use Cargo workspaces when building Soroban
projects. Let's make it easy to copy-paste these examples into their
projects, rather than requiring special tooling
(like the old `stellar contract init` behavior) to update the contract's 
Cargo.toml.

Currently, there are two compilation errors when attempting to run tests:

``` error[E0433]: failed to resolve: use of undeclared crate or module
`__get_allowance`
 --> token/src/contract.rs:67:12
  | 67 |     pub fn get_allowance(e: Env, from: Address, spender: Address)
-> Option<AllowanceValue> {
  |            ^^^^^^^^^^^^^ use of undeclared crate or module
`__get_allowance`
  | help: there is a crate or module with a similar name
  | 67 |     pub fn __allowance(e: Env, from: Address, spender: Address) ->
Option<AllowanceValue> {
  |            ~~~~~~~~~~~

error[E0223]: ambiguous associated type
 --> bls_signature/src/lib.rs:78:18
  | 78 |         agg_sig: Self::Signature,
  |                  ^^^^^^^^^^^^^^^ help: use fully-qualified syntax:
`<IncrementContractArgs as
soroban_sdk::testutils::ed25519::Sign>::Signature`
```

### What

[TODO: Short statement about what is changing.]

### Why

### Known limitations

[TODO or N/A]


chadoh:
As discussed in #328, we have decided not to implement this change.

PR #350, opened 2024-12-19
Proof of concept groth16 verifier

jayz22:
### What

See README

### Why

### Known limitations

[TODO or N/A]


groth16_verifier/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[contractimpl]
+impl Groth16Verifier {
+    pub fn verify_proof(env: Env, vk: VerificationKey, proof: Proof, pub_signals: Vec<Fr>) -> bool {
+        let bls = env.crypto().bls12_381();
+
+        // vk_x = IC0 + pub_signals[0]*IC1
+        // Here we have only one public input. If there were more, we'd sum them similarly.
+        let mut vk_x = vk.ic.get(0).unwrap();
+        if let Some(s) = pub_signals.get(0) {
```
leighmcculloch, line 37: Does the API usage to get the first and ignore the rest present an API trap to developers copying this example?

Would it be safer for contracts like this if we instead converted the `vk.ic` `Vec` into an array of length `1`, or even better a one element tuple? That should, I think, ensure that the passed in `Vec` was length `1` (please double check that if you rely on it).

leighmcculloch (commented):
One comment relating to potential API traps and if there's an opportunity to improve either the example of the SDK. I'm not the best person to review the groth specifics. Do we have anyone else who is up on that? Otherwise I can approve.

groth16_verifier/src/lib.rs
jayz22: Good point. This was an initial over-simplified version, fixed it to account for generic number of public inputs.

groth16_verifier/data/auxiliary/multiplier2.wasm
leighmcculloch, line 1: What does this wasm do or for?

groth16_verifier/README
```diff
@@ -0,0 +1,28 @@
+# Groth16 Verifier Contract
+
+A demonstration of a Groth16 zero-knowledge proof verifier implemented as a Soroban smart contract.
+
+The proof and verification key are generated following the Circom2 [getting-started guide](https://docs.circom.io/getting-started/installation/).
```
leighmcculloch, line 5: Could we mention the version of Circom2 so that folks can reproduce the same inputs?

groth16_verifier/data/auxiliary/multiplier2.wasm
jayz22: Wasm is one of the several formats the circuit is compiled to. In the circom tutorial I followed it generated the witness from the circuit wasm file (https://docs.circom.io/getting-started/computing-the-witness/#computing-the-witness-with-webassembly). So I just kept the wasm file (the intermediate artifact) here for reproducibility. 
But I guess it's not really needed since it can be produced from recompiling the `.circom` file. 
So removed it. Thanks for the catch.

groth16_verifier/README
jayz22: Done.

PR #354, opened 2025-01-16
increment_with_fuzz: fix

leighmcculloch:
Fix the increment_with_fuzz example fuzz test code that won't compile.


sisuresh (approved):
Thanks!

PR #358, opened 2025-01-19
Simplify Liquidity pool 

### LINKED ISSUE #357: simplify liquidity pool
The liquidity pool example is important as reference for devs coming in from other ecosystems and are familiar with constant-product pools. However, it seems a bit cumbersome and so I would consider some simplifications:
- [ ] Remove the use of a third part token contract for shares, this seems to complicate the code and requires manual installation of a wasm. Instead, consider maintaining shares as part of the contract's storage - no need for these to be transferable 
- [ ] Remove the enforced canonical order for the assets. For the sake of this example it doesn't matter that both liqPool(A,B) and liqPool(B,A) can coexist

Anything else that can be simplified?

tomerweller:
Update dependencies, remove canonical asset order, remove shares token contract
Close #357


liquidity_pool/src/lib.rs
```diff
@@ -92,20 +81,20 @@ fn put_reserve_b(e: &Env, amount: i128) {
...
+fn burn_shares(e: &Env, from: &Address, amount: i128) {
+    let current_shares = get_shares(e, from);
+    if current_shares < amount {
+        panic!("insufficient shares");
+    }
     let total = get_total_shares(e);
-    let share_contract = get_token_share(e);
-
-    token::Client::new(e, &share_contract).burn(&e.current_contract_address(), &amount);
+    put_shares(e, from, current_shares - amount);
```
leighmcculloch, line 90: There's no math safety around the modification of the current shares. Is that out of scope for the type of example this is trying to be?

leighmcculloch (commented):
This looks good, one concern inline.

Also, the advantage of the shares being a token is they can then be traded or moved as well, otherwise the shares are stuck as owned by a particular address. Hmm.

liquidity_pool/src/lib.rs
tomerweller: By "math safety" you mean a check that this is non negative? for `burn_shares()` or for `put_shares()`? `burn_shares()` panics if `current_shares` < `amount`

`put_total_shares()` in the current example doesn't include any checks, so I followed that pattern for `put_shares`

leighmcculloch: Presumably that safety would be provided by the share token contract though.

tomerweller: `put_total_shares()` in the current unmodified version is unchecked and does not touch the external token contract. See [here](https://github.com/stellar/soroban-examples/blob/9efe5065e644f4e4603d6a92cb6df5aa5508a483/liquidity_pool/src/lib.rs#L83).

But maybe I'm not understanding you. Can you clarify what check you'd like to see and where?

leighmcculloch: I assumed there was overflow safety in the token impl, but actually there isn't, so I don't think anything is lost here.

leighmcculloch (approved):

tomerweller:
@leighmcculloch sorry I missed your top level comment. Having pool shares represented as fungible transferable tokens is not a hard requirement for a liquidity pool. For example stellar's canonical AMM pool shares are not transferable and uniswap v3 represents positions as NFTs, not fungible tokens

PR #363, opened 2025-02-24
Demo token implementation supporting muxed address sketch.

dmkozh:
This just emits the muxed addresses in the events. The tests also show how the regular token client can be used with this.

### DIFF
--- a/token/src/test.rs
+++ b/token/src/test.rs
@@ -1,11 +1,13 @@
 #![cfg(test)]
 extern crate std;
 
+use core::ops::Add;
+
 use crate::{contract::Token, TokenClient};
 use soroban_sdk::{
     symbol_short,
-    testutils::{Address as _, AuthorizedFunction, AuthorizedInvocation},
-    Address, Env, FromVal, IntoVal, String, Symbol,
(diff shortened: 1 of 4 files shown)


token/src/test.rs
```diff
@@ -21,16 +23,85 @@ fn create_token<'a>(e: &Env, admin: &Address) -> TokenClient<'a> {
...
+        file = "target/wasm32-unknown-unknown/release/soroban_token_contract.wasm"
+    );
+}
+
+mod mux_client {
+    soroban_sdk::contractimport!(
+        file = "target/wasm32-unknown-unknown/release/soroban_token_contract.wasm",
+        expose_muxed_addresses = true
+    );
+}
```
leighmcculloch, line 37: I think the fact that every contract has two APIs, is a reasonably complex narrative.

dmkozh: This is just an option, we don't have to go this way. I just think it works fine for hiding unnecessary details 99% of the time and provides the flexibility 1% of the time this is actually necessary, similar to some other trickier concepts like non-rooted auth. But with a separate function I agree there is not much point in doing this.

PR #367, opened 2025-02-27
Handle when rustup is not in the GitPod image

leighmcculloch:
### What
Add failure tolerance to rustup uninstall command in Gitpod Dockerfile.

### Why
Prevents build failures when rustup is not already installed in the GitPod base image, which seems to change over time. It recently seemed to be removed calling an image build failure:
- https://github.com/stellar/soroban-examples/actions/runs/13535208424/job/37825492942?pr=366


leighmcculloch:
This fix was slurped into:
- https://github.com/stellar/soroban-examples/pull/366

PR #368, opened 2025-03-12
Bump cli version

Ifropc:
### What

Bump version of cli to latest

### Known limitations
N/A

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -79,7 +79,7 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -79,9 +79,7 @@ jobs:
     - run: cargo version
     - run: rustup target add ${{ matrix.sys.target }}
     - run: rustup target add wasm32-unknown-unknown
-    - uses: stellar/stellar-cli@v22.1.0
-      with:
-        version: 22.1.0
+    - uses: stellar/stellar-cli@a06ff9fd2bcd9e45a86ec3518b1a83bde898b32a
```
leighmcculloch, line 82: Why a hash here instead of the version?

Ifropc: This is temporary, I was playing around with the action from stellar-cli.
Will fix one I re-tag the cli release

PR #378, opened 2025-08-12
Update target and rust version in rust-toolchain

elizabethengelman:
### What

Update wasm target and rust version in rust-toolchain.toml

### Why
 
The e2e tests in the system-test repo clone soroban-examples and to build and test. As of https://github.com/stellar/system-test/pull/141 we are expecting contracts to be built with a target of `wasm32v1-none` and expecting the wasm to be put in a `target/wasm32v1-none/release` dir, however this repo still builds contracts with `wasm32-unknown-unknown` and puts them in a `target/wasm32-unknown-unknown/release` dir.

### Known limitations

There are other places in this repo that mention wasm32-unknown-unknown that likely should change as well.

### DIFF
--- a/account/Makefile
+++ b/account/Makefile
@@ -7,7 +7,7 @@ test: build
 
 build:
 	stellar contract build
-	@ls -l target/wasm32-unknown-unknown/release/*.wasm
+	@ls -l target/wasm32v1-none/release/*.wasm
 
 fmt:
 	cargo fmt --all
--- a/account/Makefile
+++ b/account/Makefile
@@ -7,7 +7,12 @@ test: build
 
 build:
 	stellar contract build
-	@ls -l target/wasm32v1-none/release/*.wasm || ls -l target/wasm32-unknown-unknown/release/*.wasm
+	@RUST_VERSION=$$(rustc --version | cut -d' ' -f2); \
+	if [ "$$(printf '%s\n%s\n' $$RUST_VERSION 1.81.0 | sort -V | head -n1)" = "1.81.0" ]; then \
+		ls -l target/wasm32v1-none/release/*.wasm; \
+	else \
+		ls -l target/wasm32-unknown-unknown/release/*.wasm; \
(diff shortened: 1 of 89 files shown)


fnando (approved):

account/Makefile
```diff
@@ -7,7 +7,12 @@ test: build
 
 build:
 	stellar contract build
-	@ls -l target/wasm32-unknown-unknown/release/*.wasm
+	@RUST_VERSION=$$(rustc --version | cut -d' ' -f2); \
+	if [ "$$(printf '%s\n%s\n' $$RUST_VERSION 1.81.0 | sort -V | head -n1)" = "1.81.0" ]; then \
+		ls -l target/wasm32v1-none/release/*.wasm; \
+	else \
+		ls -l target/wasm32-unknown-unknown/release/*.wasm; \
+	fi
```
elizabethengelman, line 15: I ended up needing to make this change to all of the Makefiles - I'm not sure if that's the best solution though. The issue was that each of the Makefiles has a line to presumably check to make sure that the wasm file is in the location we expect. The rust tests all call `make test`, and since we're running tests for both the up-to-date rust version and the msrv we need to be able to support both wasm locations here. 

Alternatively, I think that we could remove the check that the wasm file is in the correct dir - it seems like it was there to fail quickly and not get to the `cargo test` part. Maybe that is a cleaner solution?

elizabethengelman: Making 1.85 the minimum supported version sounds great - I was also running into an issue where a few of the other contracts needed soroban-sdk@22.0.8, but was seeing errors that that doesn't support rustc 1.80.0.
 
And updating the contract build call to add a --out-dir sounds great.

leighmcculloch: If updating to 1.85 resolves the issue with minimal otehr cahnges I'd leave the --out-dir idea, just because while it is convenient, it makes the build process work differently to regular Rust projects.

PR #396, opened 2026-01-27
Add BN254 and BLS12-381 Groth16 verifier (Circom, Gnark, and Arkworks)

mysteryon88:
### What

Added usage examples for the Soroban Groth16 verifier contract generator implemented in [this project](https://github.com/mysteryon88/soroban-verifier-gen), using proofs and verification keys produced by Circom, Gnark, and Arkworks.

### Why

These examples demonstrate how the generator can be used with popular ZK frameworks to generate Soroban-compatible verifier contracts and make the intended workflow clearer for users.

### Known limitations

- The implementation is based on the Soroban [Groth16 verifier example](https://github.com/stellar/soroban-examples/tree/main/groth16_verifier) from the Stellar repository  
- The generator and resulting contracts have not undergone a security audit


mysteryon88:
@jayz22, hi! Is there any chance that my PR will be accepted?

jayz22:
> @jayz22, hi! Is there any chance that my PR will be accepted?

@mysteryon88 sorry for the late response. 
This PR is still quite far off structurally. The two main values of this PR (as I understand it) 1. separate verification key bytes generated from separate tooling (with examples) 2. adding bn254 verifier.
So I suggest the following:
- Make two groth16 contracts, one for each curve
- Each contract should accept a verification key as input, or a `set_verification_key` method (with auth required). 
- Under `tests/` for each contract, make each tooling generated data a separate instance.
So roughly the following directory structure:
```
  groth16_verifier/
  ├── Cargo.toml              # workspace
  ├── Makefile                # iterates over contracts/
  ├── README.md
  └── contracts/
      ├── bls12_381_verifier/
      │   ├── Cargo.toml
      │   ├── Makefile
      │   ├── src/lib.rs
      │   └── tests/
      │       ├── circom.rs           # integration test
      │       ├── gnark.rs
      │       ├── arkworks.rs
      │       ├── common/mod.rs       # shared JSON parsers
      │       └── data/
      │           ├── circom/
      │           │   ├── README.md
      │           │   ├── verification_key.json
      │           │   ├── proof.json
      │           │   ├── public.json
      │           │   └── auxiliary/   # circom + snarkjs scripts
      │           ├── gnark/...
      │           └── arkworks/...
      └── bn254_verifier/
          ├── Cargo.toml
          ├── Makefile
          ├── src/lib.rs
          └── tests/
              ├── gnark.rs
              ├── common/mod.rs
              └── data/gnark/...
```
let me know if this makes sense.
(if it's and okay with you, I'm happy to tag a commit to your branch so that we can get this through the finish line sooner.)

mysteryon88:
@jayz22  thanks, this makes sense.

I’ll start restructuring the PR and should have the main changes pushed in the next couple of hours. After that, feel free to commit directly to my branch if you want to adjust or finish anything.

mysteryon88:
@jayz22  hi, I’ve done it. Could you take a look and make any edits if needed? I think it turned out pretty well.

mysteryon88:
Hi @jayz22, just a quick ping on this. Could you take a look when you have a chance?

mysteryon88:
@jayz22, I’d like to flag a security vulnerability in the current Groth16 verifier input pattern that makes this PR more time-sensitive.

While testing the verifier input boundary, I found that non-canonical public inputs can be reduced modulo the scalar-field order during conversion to `Fr`. As a result, two distinct raw 32-byte values, `x` and `x + r`, can be interpreted as the same field element and accepted with the same proof.

This does not break Groth16 soundness by itself. The vulnerability becomes exploitable when an application stores or compares the raw value, such as a nullifier or authorization identifier, while the verifier checks only its reduced field representation.

I prepared reproducible PoCs for both BN254 and BLS12-381, including a Privacy Pools example where the same proof can be reused for a second withdrawal by submitting `n + r`:

https://github.com/zk-examples/soroban-modulo-poc

I have already fixed the issue in my branch by accepting raw 32-byte public inputs, enforcing `value < r` before converting them to `Fr`, and adding regression tests for both curves.

The current example in `main` still exposes the unsafe input pattern, so users may copy it into real contracts. Could we please prioritize reviewing and merging this PR?

If the full PR is too large to merge quickly, I can immediately split the canonical-input validation and regression tests into a small security-only PR so the vulnerable pattern can be fixed first.

groth16_verifier/contracts/bls12_381_verifier/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        admin.require_auth();
+        env.storage()
+            .instance()
+            .set(&DataKey::VerificationKey, &verification_key);
+    }
+
+    pub fn verify_proof(
+        env: Env,
+        proof: Proof,
+        public_inputs: Vec<BytesN<32>>,
```
jayz22: Let's revert back to the previous interface where the verification key is passed in (and remove the `__constructor` and the admin and `set_verification_key`).
In production setting, the verification key is set once or hardcoded per contract instance. However, I would like to keep the example simple, by making it stateless and avoiding additional machinery like instance storage and auth, which has been demonstrated elsewhere. (This contract is also used by another example - privacy pool, which would also have to adapt to the new requirements). 
Since you moved the repo structure, please also update the groth16 contract link in `privacy_pool` example.

jayz22: please change
`public_inputs: Vec<BytesN<32>>` to
`public_inputs: Vec<Bls12381Fr>`.

I'm not that concerned about the modular wrapping issue. 
`Bls12381Fr` is s a typed scalar that makes sure modulo is applied during its construction. So the caller to the contract need to take care of the modular wrapping such that within the contract there is no ambiguity.

Same for the BN254 contract, use `Bn254Fr`.

groth16_verifier/contracts/bls12_381_verifier/tests/common/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        Some(values) => values,
+        None => {
+            let public_path = dir.join("public.json");
+            serde_json::from_str(&read_file(public_path)).unwrap()
+        }
+    };
+
+    let mut signals = Vec::new(env);
+    for signal in public_signals {
+        signals.push_back(fr_bytes_from_str(env, &signal));
```
jayz22: if the public signal type is `Bls12381Fr` (see my comment above), then in order to construct it, you will be applying modulo. Here you can assert the input being lesser than the modular before the construction, if you want.

jayz22 (changes requested):

jayz22:
@mysteryon88 Thanks for adapting the PR to my earlier comments. The repo structure looks good now. I've left a couple of comments regarding simplifying the contract and the modular wrapping issue you mentioned above. Once addressed we can merge it.

mysteryon88:
@jayz22 Thanks for the review. I’ve addressed all the requested changes, and everything should now be ready for another review. Thank you!

jayz22 (approved):
Looks good, thanks for the back and forth!

PR #406, opened 2026-04-08
Replace math operators with checked/strict alternatives in contract examples

gabrielrondon:
### What
Replace bare arithmetic operators (`+`, `-`, `*`, `/`, `+=`) with their checked equivalents (`checked_add`, `checked_sub`, `checked_mul`, `checked_div`) across 12 contract example source files.

### Why
Encourage developers to think about overflow failure cases when copying these examples as starting points for their own contracts. Raw arithmetic operators silently wrap or panic on overflow — checked functions make overflow handling explicit.

Ref: #383

### Changed examples
- `increment`, `events`, `errors`, `auth`, `bls_signature`, `increment_with_pause`, `increment_with_fuzz`, `other_custom_types` — simple counter increments
- `custom_types` — parameterized increment
- `alloc` — accumulator loop
- `liquidity_pool` — all pool arithmetic (deposit, swap, withdraw, mint, burn)
- `mint-lock` — epoch calculations

### Not changed (by design)
- Test files — test arithmetic is fine with operators
- `privacy-pools` — ZK/BLS math internals where checked ops would add noise
- Fuzzing targets — arbitrary generation uses range arithmetic

### Known limitations
N/A

### DIFF
--- a/alloc/src/lib.rs
+++ b/alloc/src/lib.rs
@@ -13,9 +13,9 @@ impl AllocContract {
         let mut v1 = alloc::vec![];
         (0..count).for_each(|i| v1.push(i));
 
-        let mut sum = 0;
+        let mut sum: u32 = 0;
         for i in v1 {
-            sum += i;
+            sum = sum.checked_add(i).expect("overflow");
         }
 
         sum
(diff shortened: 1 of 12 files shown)


gabrielrondon:
Friendly ping. This has been open since April with only the Copilot review, and it still merges cleanly into main.

The motivation stands: these examples are the first thing new contract developers copy, and bare arithmetic in them normalizes overflow-prone patterns in real contracts. Happy to rebase or narrow the scope (for example, only the counter-style examples) if that makes review easier.

PR #410, opened 2026-07-23
Disable resource limits in bn254 wasm test

### LINKED ISSUE #1931: `InvocationResourceLimits::mainnet()` is out of date with actual Mainnet network settings
### What version are you using?

soroban-sdk 27.0.0 (current `main`, e5cb4b52). The values appear to date back to #1677, which introduced resource limit enforcement.

### What did you do?

Compared the hardcoded limits in [`InvocationResourceLimits::mainnet()`](https://github.com/stellar/rs-soroban-sdk/blob/main/soroban-sdk/src/testutils/cost_estimate.rs), which `Env::default()` has enforced on every test invocation since v25, with live Mainnet settings fetched on 2026-07-10 (`stellar network settings --network mainnet`; also visible on [Stellar Lab: Network Limits](https://lab.stellar.org/network-limits)).

### What did you expect to see?

The SDK limits should match Mainnet's per-transaction limits, since the enforcement exists to warn developers when a contract "might be too resource-heavy to run on a real network".

### What did you see instead?

These fields differ:

- `instructions`: the SDK uses 600,000,000 instead of 400,000,000
- `disk_read_entries`: the SDK uses 100 instead of 200
- `write_entries`: the SDK uses 50 instead of 200
- `ledger_entries`: the SDK uses 100 instead of 400

The remaining fields match: `mem_bytes`, `disk_read_bytes`, `write_bytes`, `contract_events_size_bytes`, `max_contract_data_key_size_bytes`, `max_contract_data_entry_size_bytes`, `max_contract_code_entry_size_bytes`.

The two directions of staleness have different consequences:

- **`instructions` too high:** a contract can pass `Env::default()` tests yet exceed Mainnet's transaction limit in production, defeating the enforcement's early-warning purpose.
- **Entry limits too low:** tests can panic for contracts that are valid on Mainnet, pushing developers toward `disable_resource_limits()` and removing the safety net entirely.

### Why is this important?

AI coding assistants often treat SDK values and rustdocs as ground truth for network behavior. I have seen assistants rely on these stale limits when judging whether a contract fits within Mainnet's budget, misleading both the model and the developers using it.

### Related staleness in the same file

`CostEstimate::fee()` uses a Pubnet fee snapshot from 2024-12-11 that could be refreshed in the same pass:

| Fee | SDK snapshot | Mainnet (2026-07-10) |
|---|---:|---:|
| `fee_per_instruction_increment` | 25 | 7 |
| `fee_per_disk_read_entry` | 6,250 | 1,563 |
| `fee_per_write_entry` | 10,000 | 2,500 |
| `fee_per_disk_read_1kb` | 1,786 | 447 |
| `fee_per_write_1kb` | 3,500 | 875 (`fee_write1_kb`) |
| `fee_per_historical_1kb` | 16,235 | 4,059 |
| `fee_per_contract_event_1kb` | 10,000 | 5,000 |
| `fee_per_transaction_size_1kb` | 1,624 | 406 |

### Suggested fix

1. Update the `mainnet()` rustdocs to point readers — human and AI alike — to a live source (`stellar network settings --network mainnet` or [Stellar Lab: Network Limits](https://lab.stellar.org/network-limits)) rather than relying on the hardcoded snapshot.
2. Update `mainnet()` to the current values (`instructions: 400_000_000`, `disk_read_entries: 200`, `write_entries: 200`, `ledger_entries: 400`).
3. Refresh the `fee()` snapshot and its "as of" comment.
4. Consider a scheduled CI job, or a release-checklist item, that diffs `mainnet()` against `stellar network settings --network mainnet` — nothing currently catches drift. Since Mainnet limits are validator-voted and change infrequently, a weekly or monthly run would be enough.

leighmcculloch:
### What

Disable resource limits in the `import_ark_bn254` example.

### Why

`Env::default()` in tests enforces mainnet limits. The refresh in stellar/rs-soroban-sdk#1946 (issue stellar/rs-soroban-sdk#1931) updates those limits to match current mainnet, lowering the per-transaction instruction ceiling from 600,000,000 to the true value of 400,000,000. This example consumes ~404,556,039 instructions, so under the refreshed ceiling it panics with `HostError: Error(Budget, ExceededLimit)`.

### Known limitations

N/A

---
_Generated by [Claude Code](https://claude.ai/code/session_01Grhpc1bfnRpvtKtTK5Y2EC)_

### DIFF
--- a/import_ark_bn254/src/test.rs
+++ b/import_ark_bn254/src/test.rs
@@ -61,9 +61,11 @@ mod bn254_contract {
 #[test]
 fn test_running_contract_as_wasm() {
     let env = Env::default();
-    // This contract is too expensive to run with production budget, let's first
-    // reset the budget to unlimited
+    // This contract is too expensive to run with production budget and mainnet
+    // resource limits, so first reset the budget to unlimited and disable the
+    // resource limit enforcement.
     env.cost_estimate().budget().reset_unlimited();
+    env.cost_estimate().disable_resource_limits();
 
     let contract_id = env.register(bn254_contract::WASM, ());
     let client = bn254_contract::Client::new(&env, &contract_id);
--- a/import_ark_bn254/src/test.rs
+++ b/import_ark_bn254/src/test.rs
@@ -64,7 +64,6 @@ fn test_running_contract_as_wasm() {
     // This contract is too expensive to run with production budget and mainnet
     // resource limits, so first reset the budget to unlimited and disable the
     // resource limit enforcement.
-    env.cost_estimate().budget().reset_unlimited();
     env.cost_estimate().disable_resource_limits();
 
     let contract_id = env.register(bn254_contract::WASM, ());
--- a/import_ark_bn254/src/test.rs
+++ b/import_ark_bn254/src/test.rs
@@ -64,6 +64,7 @@ fn test_running_contract_as_wasm() {
     // This contract is too expensive to run with production budget and mainnet
     // resource limits, so first reset the budget to unlimited and disable the
     // resource limit enforcement.
+    env.cost_estimate().budget().reset_unlimited();
     env.cost_estimate().disable_resource_limits();
 
     let contract_id = env.register(bn254_contract::WASM, ());


leighmcculloch:
The failing `build-and-test (account, latest, …)` check is a transient CI infrastructure error, not a problem with this change. The job died during setup while downloading the `stellar-cli` binary — `curl: (35) Recv failure: Connection reset by peer` — before anything compiled or ran. This diff only touches `import_ark_bn254/src/test.rs`, so it can't affect the `account` example. It should clear on a re-run of the failed job.

---
_Generated by [Claude Code](https://claude.ai/code)_

import_ark_bn254/src/test.rs
```diff
@@ -61,9 +61,11 @@ mod bn254_contract {
 #[test]
 fn test_running_contract_as_wasm() {
     let env = Env::default();
-    // This contract is too expensive to run with production budget, let's first
-    // reset the budget to unlimited
+    // This contract is too expensive to run with production budget and mainnet
+    // resource limits, so first reset the budget to unlimited and disable the
+    // resource limit enforcement.
     env.cost_estimate().budget().reset_unlimited();
+    env.cost_estimate().disable_resource_limits();
```
mootz12: why does `reset_unlimited` not work here? Is this the shadow CPU budget overflowing / not getting reset during `reset_unlimited`?

This seems like it shouldn't be necessary.

mootz12 (approved):

import_ark_bn254/src/test.rs
leighmcculloch: It's not the shadow budget — `reset_unlimited()` and the resource-limit check act on two independent mechanisms.

`budget().reset_unlimited()` raises the metering **budget's ceilings** to `u64::MAX`, which is what stops execution itself from tripping a `Budget, ExceededLimit` while the pairing runs.

The mainnet check is separate: since v25, `Env::default()` enables the host's `InvocationMeter` with `InvocationResourceLimits::mainnet()`. For each top-level invocation the meter resets the budget's *consumed* counters, runs the call, then reads back `get_cpu_insns_consumed()` and calls `verify_limits()`, panicking if `instructions > limits.instructions`. That comparison is against the stored mainnet limit, not against the budget's ceiling — so raising the ceiling with `reset_unlimited()` doesn't affect it. `disable_resource_limits()` is the only knob that turns that check off (`set_invocation_resource_limits(None)`).

So both are needed here, and it's expected rather than a bug. I confirmed it empirically against rs-soroban-sdk#1946's branch: with only `reset_unlimited()` the wasm test still panics with `Error(Budget, ExceededLimit)` — `invocation resource limits are exceeded: instructions: 404556039 > 400000000` — and adding `disable_resource_limits()` makes it pass. The `as_native` test stays untouched because it never crosses the wasm/invocation boundary that the meter measures.

---
_Generated by [Claude Code](https://claude.ai/code)_

dmkozh: While `disable_resource_limits` is necessary here, I don't think we should need `budget().reset_unlimited()` (we can't really hide it, but at least we can remove it from our examples). Invocation metering is enabled by default, so resetting the budget explicitly won't work most of the time.

leighmcculloch: I tried removing it, and the wasm test then panics with `Error(Budget, ExceededLimit)` at the `mock_verify` call itself (host.rs:910) — on both the current released SDK (27.0.0) and the #1946 branch (27.0.1), so it still seems required today.

Tracing it through the host: the per-invocation reset in `InvocationMeter::push_invocation` calls `budget.reset()`, which clears the *consumed* counters (`BudgetTracker::reset`) but leaves the configured ceilings in place — it doesn't restore the default limit. So the `u64::MAX` ceiling from `reset_unlimited()` does carry into the metered invocation (the budget dump shows `Cpu limit: 18446744073709551615; used: 560718778`), and without it the call runs under the default ~100M CPU ceiling, so the ~560M-instruction pairing trips the budget before the invocation-limit check is ever reached.

Is the "won't work most of the time" behavior something coming in a newer env-host — e.g. the meter running the top-level call under an unlimited budget and gating purely via the invocation limits? On the versions in play, dropping the line fails, so I've kept it for now. Happy to remove it if there's a host change that makes it unnecessary, or to take a different approach if you'd prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

leighmcculloch: Apologies for the agent comments here, that was unintentional. I enabled the CI autofix feature on a few PRs and I have not seen it eagerly respond to comments until now. It's unclear to me if that is rogue behaviour or intended behaviour.

leighmcculloch: > I don't think we should need `budget().reset_unlimited()`

It seems both calls are necessary. The test API seems to have become a bit muddled. I'll open an issue to review the API and figure out what we can change, because having to call both is not intuitive and even after discovering you have to call both is confusing to me.

In the meantime we should merge this with using both here since that is what is required today.

dmkozh: Yeah, that's what I meant - we need to fix `disable_resource_limits` to also make the budget unlimited.

leighmcculloch: Follow up:
- https://github.com/stellar/rs-soroban-sdk/issues/1949

PR #415, opened 2026-09-09
Rename token example's approve arg to live_until_ledger

leighmcculloch:
### What

Rename the `approve` allowance parameter and its internal storage field in the token example from `expiration_ledger` to `live_until_ledger`.

### Why

The soroban-sdk v27.0.1 (stellar/rs-soroban-sdk#1932) renamed this argument on both `TokenInterface` and `StellarAssetInterface` to align with the SEP-41 spec and this is just keeping the names updated everywhere we can.

The event field still has the old name because that was updated in v28 of the sdk, so we can't update it yet here where the examples are still on v27.

### Known limitations

N/A


dmkozh (approved):

leighmcculloch:
Follow up in the stellar-docs:
- https://github.com/stellar/stellar-docs/pull/2846

